PR #1, opened 2022-01-14
feat: [CO-89] user attribute cannot be part of their password

keshavbhatt:
- added new password policy which will prevent user from adding their
personal data in their passowrd string
- this policy is enforced by default and can be set to false in COS
or per user like other password restriction that are already in place

### DIFF
--- a/store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
+++ b/store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
@@ -5,33 +5,6 @@
 
 package com.zimbra.cs.account.ldap;
 
-import java.io.IOException;
-import java.io.PrintWriter;
-import java.security.SecureRandom;
-import java.util.ArrayList;
-import java.util.Arrays;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Comparator;
-import java.util.Date;
-import java.util.HashMap;
-import java.util.HashSet;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Random;
-import java.util.Set;
-import java.util.Stack;
-import java.util.TreeMap;
-import java.util.TreeSet;
-import java.util.regex.Pattern;
-import java.util.regex.PatternSyntaxException;
-import java.util.stream.Collectors;
-
-import java.util.stream.Stream;
-import org.apache.commons.codec.binary.Hex;
-import org.apache.commons.lang.StringUtils;
-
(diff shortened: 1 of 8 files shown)


store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -6219,6 +6240,42 @@ private void checkPasswordStrength(String password, Account acct, Cos cos, ZMuta
...
+            .toArray(String[]::new);
+
+    for (String part : dictionary) {
+      if (part.length() > 2 && (part.contains(passwordLower) || passwordLower.contains(part))) {
+        throw AccountServiceException.INVALID_PASSWORD(
+            "password contains username or other personal data: " + part,
+            new Argument(
+                Provisioning.A_zimbraBlockPersonalDataInPasswordEnabled, part, Argument.Type.STR));
+      }
+    }
```
Polpetta, line 6276: You can execute all this code in the same stream for better performances:
```suggestion
    Optional<String> matchedSensitivePart =
        Stream.of(personalDataImploded, cleanEmailExploded, fullNameExploded)
            .flatMap(Stream::of)
            .parallel()
            .anyMatch(part -> part.length() > 2 && (part.contains(passwordLower) || passwordLower.contains(part))))
            .findAny();

    if (matchedSensitivePart.isPresent()) {
        throw AccountServiceException.INVALID_PASSWORD(
            "password contains username or other personal data: " + part,
            new Argument(
                Provisioning.A_zimbraBlockPersonalDataInPasswordEnabled, part, Argument.Type.STR));
    }
```

Polpetta (changes requested):
Please fix Javadoc & check out stream solution

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -6219,6 +6240,42 @@ private void checkPasswordStrength(String password, Account acct, Cos cos, ZMuta
         }
     }
 
+  // validate that user is not using personal data in his password
+  private void validatePasswordEntropy(String password, Account acct)
```
Polpetta, line 6244: Please also add some unit testing for this method since it seems to be quite important and I don't see tests

Polpetta:
Mmmm why you closed this? :thinking: Could have been simpler to keep everything in a single PR instead of opening a new one. Now I have to switch tab to check what I wrote here and see if you fixed it :/

PR #2, opened 2022-01-17
Feat/service discover expose

frisonisland:
Feature: add service discovery integration for mailbox.

### DIFF
--- a/packages/appserver-service/PKGBUILD
+++ b/packages/appserver-service/PKGBUILD
@@ -33,11 +33,11 @@ priority="optional"
 package() {
   cd "${srcdir}"/../../staging
   install -D packages/appserver-service/carbonio-mailbox-sidecar.service \
-  "${pkgdir}/lib/systemd/system/carbonio-mailbox-sidecar.service"
+    "${pkgdir}/lib/systemd/system/carbonio-mailbox-sidecar.service"
   install -D packages/appserver-service/carbonio-mailbox.hcl \
-  "${pkgdir}/etc/zextras/service-discover/carbonio-mailbox.hcl"
+    "${pkgdir}/etc/zextras/service-discover/carbonio-mailbox.hcl"
   install -D packages/appserver-service/carbonio-mailbox \
-  "${pkgdir}/usr/bin/carbonio-mailbox"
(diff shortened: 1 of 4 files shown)


M0Rf30 (approved):

keshavbhatt (approved):

M0Rf30 (approved):

packages/appserver-service/PKGBUILD
```diff
@@ -20,6 +20,8 @@ depends=(
   "carbonio-appserver-conf"
   "carbonio-appserver-war"
   "carbonio-appserver"
+  "service-discover"
+  "pending-setups"
```
Polpetta, line 24: Note coordinate with @davide-baldo or @ZeXtrasJay for the name, since for what I understand `service-discover` maybe not the final name. Same for `pending-setups`

Polpetta (changes requested):
Check the name with our Architects, then the rest lgtm

packages/appserver-service/PKGBUILD
frisonisland: Hey @davide-baldo , thanks for the feedback.
I'll add a service-intention kind definition for the user-management service.

About final naming I think we're all set at this point :)

davide-baldo (approved):

Polpetta (approved):
lgtm!

Polpetta:
@frisonisland please merge commits always following conventional commit convention: https://www.conventionalcommits.org/en/v1.0.0/

frisonisland:
@Polpetta sorry for the inconvenience, thought I wrote "feat" instead of "Feature".

frisonisland:
@davide-baldo sorry for pointing it out now, but I think user-management definition is wrong and the service name should be carbonio-user-management. Can you confirm that? I'll open another PR.

davide-baldo:
> @davide-baldo sorry for pointing it out now, but I think user-management definition is wrong and the service name should be carbonio-user-management. Can you confirm that? Other PR: #6

Yep, it's carbonio-user-management, my bad not using the full form, check out of the service-configurations document in dev guide

PR #3, opened 2022-01-17
bugfix/co-152 memcached binding

flanciotti:
Memcached binding address is now correctly read from settings. Nginx uses Memcached binding address configuration to connect to memcached service.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -63,16 +63,6 @@ pipeline {
             }
         }
         stage('Build deb/rpm') {
-            when {
-                anyOf {
-                    branch 'release/*'
-                    branch 'custom/*'
-                    branch 'beta/*'
-                    branch 'playground/*'
-                    expression { params.PLAYGROUND == true }
-                    buildingTag()
-                }
-            }
             stages {
                 stage('pacur') {
                     parallel {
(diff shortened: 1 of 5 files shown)


M0Rf30 (approved):

keshavbhatt (approved):

Jenkinsfile
```diff
@@ -63,16 +63,6 @@ pipeline {
             }
         }
         stage('Build deb/rpm') {
-            when {
```
Polpetta, line 66: Why did you changed this part?

Jenkinsfile
```diff
@@ -63,16 +63,6 @@ pipeline {
...
-                anyOf {
-                    branch 'release/*'
-                    branch 'custom/*'
-                    branch 'beta/*'
-                    branch 'playground/*'
-                    expression { params.PLAYGROUND == true }
-                    buildingTag()
-                }
-            }
```
Polpetta, line 75: Why did you changed this part?

M0Rf30: because otherwise it never generate packages. It also happened on nginx-conf and it always returns green light, but packaging process was wrong. Upload and promotion conditions are defererred to next stages

Polpetta: Ok. Please next time specify in the commit message. @flanciotti please adhere to https://www.conventionalcommits.org/en/v1.0.0/ when committing

PR #5, opened 2022-01-21
New Password policy without LDAP attribute

keshavbhatt:
user attribute cannot be part of their password

added new password policy which will prevent users from adding their
personal data in their password string

this policy is enforced by default

### DIFF
--- a/store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
+++ b/store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
@@ -2,30 +2,23 @@ package com.zimbra.cs.account.ldap;
 
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.cs.account.Account;
+import com.zimbra.cs.account.AccountServiceException;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.cs.mailbox.MailboxTestUtil;
-import java.util.Arrays;
 import java.util.HashMap;
-import java.util.Optional;
 import java.util.UUID;
-import java.util.stream.Stream;
-import org.junit.Assert;
 import org.junit.Before;
 import org.junit.BeforeClass;
 import org.junit.Test;
 
-
-/**
- * Unit test for {@link LdapProvisioning}.
- *
- */
+/** Unit test for {@link LdapProvisioning}. */
 public class LdapProvisioningTest {
 
   @BeforeClass
   public static void init() throws Exception {
     MailboxTestUtil.initServer();
     Provisioning prov = Provisioning.getInstance();
-    HashMap<String,Object> attrs;
+    HashMap<String, Object> attrs;
     attrs = new HashMap<>();
     attrs.put(Provisioning.A_zimbraId, UUID.randomUUID().toString());
     attrs.put(Provisioning.A_sn, "Leesa");
@@ -33,6 +26,15 @@ public class LdapProvisioningTest {
     attrs.put(Provisioning.A_initials, "James");
     attrs.put(Provisioning.A_mail, "natalie.leesa@example.com");
     prov.createAccount("natalie.leesa@example.com", "testpassword", attrs);
+
+    HashMap<String, Object> attrs2;
+    attrs2 = new HashMap<>();
+    attrs2.put(Provisioning.A_zimbraId, UUID.randomUUID().toString());
+    attrs2.put(Provisioning.A_sn, "Lamborghini");
+    attrs2.put(Provisioning.A_cn, "Milano");
+    attrs2.put(Provisioning.A_initials, "Cars");
+    attrs2.put(Provisioning.A_mail, "milano@lamborghini");
+    prov.createAccount("milano@lamborghini", "testpassword", attrs2);
   }
 
   @Before
@@ -41,58 +43,47 @@ public class LdapProvisioningTest {
   }
 
   /**
-   * This test is to validate functionality of
-   * ({@link com.zimbra.cs.account.ldap.LdapProvisioning#validatePasswordEntropyForPersonalData(String, Account)})
-   * @throws ServiceException */
-  @Test
+   * This test is to validate functionality of ({@link
+   * com.zimbra.cs.account.ldap.LdapProvisioning#validatePasswordEntropyForPersonalData(String,
+   * Account)})
+   *
+   * @throws ServiceException
+   */
+  @Test(expected = AccountServiceException.class)
   public void shouldFindPersonalDataInPassword() throws ServiceException {
 
     Account acct = Provisioning.getInstance().getAccount("natalie.leesa@example.com");
 
-    //fake passwords for test against created user account
-    String[] testPasswords = {"natalie", "Natalie123", "Leesa34", "example01", "Leesa.Example", "eXamPle", "2022james"};
-
-    // get possible personal data for account
-    Optional<String> sName =
-        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_sn)).orElse(""));
-    Optional<String> cName =
-        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_cn)).orElse(""));
-    Optional<String> initials =
-        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_initials)).orElse(""));
-    Optional<String> email =
-        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_mail)).orElse(""));
-
-    // prepare dictionary
-    String[] fullNameExploded = cName.get().toLowerCase().split(" ");
-    String[] emailExploded = email.get().split("[@._]");
-    String[] cleanEmailExploded =
-        Arrays.copyOf(emailExploded, emailExploded.length - 1); // remove the top level domain(TLD)
-    Arrays.stream(cleanEmailExploded)
-        .forEach(
-            em -> {
-              System.out.println("email element" + em);
-            });
-
-    // find match in dictionary and password
-    Arrays.stream(testPasswords).map(String::toLowerCase).forEach(passwordLower -> {
-      String[] personalDataImploded = {
-          sName.get().toLowerCase(),
-          cName.get().toLowerCase(),
-          cName.get().replace(" ", "").toLowerCase(),
-          initials.get().toLowerCase()
-      };
-      Optional<String> matchedSensitivePart =
-          Stream.of(personalDataImploded, cleanEmailExploded, fullNameExploded)
-              .flatMap(Stream::of)
-              .parallel()
-              .filter(
-                  part ->
-                      (part.length() > 2
-                          && (part.contains(passwordLower) || passwordLower.contains(part))))
-              .findAny();
-      Assert.assertTrue("matching against password:" + passwordLower,
-          matchedSensitivePart.isPresent());
-    });
+    // fake passwords for test against created user account
+    String[] testPasswords = {
+      "natalie", "Natalie123", "Leesa34", "example01", "Leesa.Example", "eXamPle", "2022james"
+    };
+    for (String testPassword : testPasswords) {
+      String passwordLower = testPassword.toLowerCase();
+      LdapProvisioning.validatePasswordEntropyForPersonalData(passwordLower, acct);
+    }
   }
-  
+
+  /**
+   * This test is to validate functionality of ({@link
+   * com.zimbra.cs.account.ldap.LdapProvisioning#validatePasswordEntropyForPersonalData(String,
+   * Account)}) for specific case in which user might be using a dotless domain
+   *
+   * @throws ServiceException
+   */
+  @Test(expected = AccountServiceException.class)
+  public void shouldFindPersonalDataInPasswordForDotLessDomainsEmails() throws ServiceException {
+
+    Account acct = Provisioning.getInstance().getAccount("milano@lamborghini");
+
+    // fake passwords for test against created user account
+    String[] testPasswords = {
+      "milano", "Lamborghini123", "Milano34", "lamborghini01", "Milano.Lamborghini", "lAmbOrgHini", "2022cars"
+    };
+
+    for (String testPassword : testPasswords) {
+      String passwordLower = testPassword.toLowerCase();
+      LdapProvisioning.validatePasswordEntropyForPersonalData(passwordLower, acct);
+    }
+  }
 }
--- a/store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
+++ b/store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
@@ -1,0 +1,99 @@
+package com.zimbra.cs.account.ldap;
+
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.account.Account;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.mailbox.MailboxTestUtil;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Stream;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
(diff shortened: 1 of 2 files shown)


store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /**
+   * This test is to validate functionality of
+   * ({@link com.zimbra.cs.account.ldap.LdapProvisioning#validatePasswordEntropyForPersonalData(String, Account)})
+   * @throws ServiceException */
+  @Test
+  public void shouldFindPersonalDataInPassword() throws ServiceException {
+
+    //Account acct = Provisioning.getInstance().get(AccountBy.id, "natalie.leesa@example.com");
```
Polpetta, line 50: You can remove this line
```suggestion
```

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+    // get possible personal data for account
+    Optional<String> sName =
+        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_sn)).orElse(""));
+    Optional<String> cName =
+        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_cn)).orElse(""));
+    Optional<String> initials =
+        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_initials)).orElse(""));
+    Optional<String> email =
+        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_mail)).orElse(""));
```
Polpetta, line 64: Why this? You can simply do this:

```suggestion
    Optional<String> sName = Optional.ofNullable(acct.getAttr(Provisioning.A_sn));
    Optional<String> cName = Optional.ofNullable(acct.getAttr(Provisioning.A_cn));
    Optional<String> initials = Optional.ofNullable(acct.getAttr(Provisioning.A_initials));
    Optional<String> email = Optional.ofNullable(acct.getAttr(Provisioning.A_mail));
```

KISS

Polpetta:
See https://github.com/Zextras/carbonio-mailbox/pull/1 for reference

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
```diff
@@ -0,0 +1,99 @@
...
+
+/**
+ * Unit test for {@link LdapProvisioning}.
+ *
+ */
+public class LdapProvisioningTest {
+
+  @BeforeClass
+  public static void init() throws Exception {
+    MailboxTestUtil.initServer();
```
Polpetta, line 26: Don't initialize stuff in the `static` field. This will lead to problems once all tests are initialized this way. If you are going to start stuff please start it in `@Before` block. It is slower I know, but Java has a hard limit for static memory entries and unfortunately by doing this you're polluting this space, resulting in a KABOM once we reach a good number of tests.

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
```diff
@@ -0,0 +1,99 @@
...
+    attrs.put(Provisioning.A_cn, "Natalie Leesa");
+    attrs.put(Provisioning.A_initials, "James");
+    attrs.put(Provisioning.A_mail, "natalie.leesa@example.com");
+    prov.createAccount("natalie.leesa@example.com", "testpassword", attrs);
+  }
+
+  @Before
+  public void setUp() throws Exception {
+    MailboxTestUtil.clearData();
+  }
```
Polpetta, line 41:
```suggestion
  @Before
  public void setUp() throws Exception {
    MailboxTestUtil.initServer();
    Provisioning prov = Provisioning.getInstance();
    HashMap<String,Object> attrs;
    attrs = new HashMap<>();
    attrs.put(Provisioning.A_zimbraId, UUID.randomUUID().toString());
    attrs.put(Provisioning.A_sn, "Leesa");
    attrs.put(Provisioning.A_cn, "Natalie Leesa");
    attrs.put(Provisioning.A_initials, "James");
    attrs.put(Provisioning.A_mail, "natalie.leesa@example.com");
    prov.createAccount("natalie.leesa@example.com", "testpassword", attrs);
  }
```

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
```diff
@@ -0,0 +1,99 @@
...
+              .parallel()
+              .filter(
+                  part ->
+                      (part.length() > 2
+                          && (part.contains(passwordLower) || passwordLower.contains(part))))
+              .findAny();
+      Assert.assertTrue("matching against password:" + passwordLower,
+          matchedSensitivePart.isPresent());
+    });
+  }
```
Polpetta, line 97: I don't understand this part of the test. Here you're supposed to call the real method, to test if it respects its contract (provided with a given input it gives back the right output). Why did you copied the production code here?

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -6219,6 +6221,60 @@ private void checkPasswordStrength(String password, Account acct, Cos cos, ZMuta
...
+      throws AccountServiceException {
+    // get possible personal data for account
+    Optional<String> sName =
+        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_sn)).orElse(""));
+    Optional<String> cName =
+        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_cn)).orElse(""));
+    Optional<String> initials =
+        Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_initials)).orElse(""));
+    Optional<String> email =
+      Optional.of(Optional.ofNullable(acct.getAttr(Provisioning.A_mail)).orElse(""));
```
Polpetta, line 6245: Check out `provisioningTest`

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
```diff
@@ -0,0 +1,99 @@
...
+import com.zimbra.cs.mailbox.MailboxTestUtil;
+import java.util.Arrays;
+import java.util.HashMap;
+import java.util.Optional;
+import java.util.UUID;
+import java.util.stream.Stream;
+import org.junit.Assert;
+import org.junit.Before;
+import org.junit.BeforeClass;
+import org.junit.Test;
```
Polpetta, line 15: I see `junit4` being used here. Can't we use `junit5` already?

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -6219,6 +6221,60 @@ private void checkPasswordStrength(String password, Account acct, Cos cos, ZMuta
...
+    String[] emailExploded = email.get().split("[@._]");
+    String[] cleanEmailExploded =
+        Arrays.copyOf(emailExploded, emailExploded.length - 1); // remove the top level domain(TLD)
+    String passwordLower = password.toLowerCase();
+    String[] personalDataImploded = {
+      sName.get().toLowerCase(),
+      cName.get().toLowerCase(),
+      cName.get().replace(" ", "").toLowerCase(),
+      initials.get().toLowerCase()
+    };
```
Polpetta, line 6258: Have you considered the case where the email is `something@example`? This is a valid email address, since top level domain can be email address. So for example, `lamborghini` has its own TLD, so `milano@lamborghini` is a valid email address (for instance https://milano.lamborghini/ is a valid website, and they have their own email address too).

Please keep this in consideration too.

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
```diff
@@ -6219,6 +6221,60 @@ private void checkPasswordStrength(String password, Account acct, Cos cos, ZMuta
...
+            .flatMap(Stream::of)
+            .parallel()
+            .filter(
+                part ->
+                    (part.length() > 2
+                        && (part.contains(passwordLower) || passwordLower.contains(part))))
+            .findAny();
+    if (matchedSensitivePart.isPresent()) {
+      throw AccountServiceException.INVALID_PASSWORD(
+          "password contains username or other personal data: " + matchedSensitivePart.get(),
```
Polpetta, line 6272: `password` here is lowercase because all the other errors are like that? Or is it `Password`?

frisonisland (approved):

Polpetta (changes requested):
Tests need to be fixed, and TLD email considered too. I'd consider the following email, that should have a password that respectively contains or not a sensitive data in it (so for each email address 2 tests cases):
* `emily@example`
* `emily@example.com`
* `emily.surname@example`
* `emily.surname@example.com`

Note that with Junit you can use Parameterized Tests to help you out with the same test on a different set of data: https://www.baeldung.com/parameterized-tests-junit-5. Please check out if it is worth doing that way.

Polpetta:
As a final note, please stick to this PR, don't close it otherwise all the history of the changes I requested will be lost and I'll have to manually keep track of them.
Push on it as much as you need, we'll squash everything anyway so only one commit will come out of this :blush:

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
keshavbhatt: Sure.

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
keshavbhatt: I don't want to deal with noSuchElementException in my code below these initializations. 
Obviously there could be 1 thousand other ways to keep it simple, but this one works for me. :)

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
keshavbhatt: I got "java.lang.Exception: Method init() should be static" error when i tried to do it without init,

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
keshavbhatt: ? In the future we might have more tests in this class and the idea is to use MailboxTestUtil.clearData(); to provide clear environment for each test.

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
keshavbhatt: I know :) that's why i used reference in java doc to that particular method. 

I cannot initialize LdapProvisioning class just to perform a test on its helper method. That's why it was done like this. 
I might follow the approach you asking for but there i would require to create a separate class with just one method.

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
keshavbhatt: What ?

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
keshavbhatt: I don't want to introduce another dependency just to write tests for this case only. I found existing junit4 that is adopted by this project everywhere sufficient enough to test my code. We might consider adding jUnit5 and rewriting all tests to achieve this goal.

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
keshavbhatt: Yes, I have considered.

Also https://www.icann.org/en/announcements/details/new-gtld-dotless-domain-names-prohibited-30-8-2013-en

keshavbhatt:
> As a final note, please stick to this PR, don't close it otherwise all the history of the changes I requested will be lost and I'll have to manually keep track of them. Push on it as much as you need, we'll squash everything anyway so only one commit will come out of this blush

I understand your concerns. It's an internal thing that forced me to create another PR. 
I understand how git works at least I know the basics :)
All feedbacks are appritiated.

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
keshavbhatt: I am now using the direct invoke to the method, at cost of making it public and static

keshavbhatt (commented):
updated the PR with all necessary changes

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
Polpetta: Ok, then keep it in the static field and let's hope for the best

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
```diff
@@ -0,0 +1,89 @@
...
+    MailboxTestUtil.clearData();
+  }
+
+  /**
+   * This test is to validate functionality of ({@link
+   * com.zimbra.cs.account.ldap.LdapProvisioning#validatePasswordEntropyForPersonalData(String,
+   * Account)})
+   *
+   * @throws ServiceException
+   */
```
Polpetta, line 51: This is not something to fix but is a tip: javadoc in tests is not mandatory! It will help people in the future to maintain this code tho

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
Polpetta: I mean I don't understand why `Optional.of(Optional.ofNullable(...).orElse(""))` when you can write this way:
```suggestion
    Optional<String> sName = Optional.ofNullable(acct.getAttr(Provisioning.A_sn));
    Optional<String> cName = Optional.ofNullable(acct.getAttr(Provisioning.A_cn));
    Optional<String> initials = Optional.ofNullable(acct.getAttr(Provisioning.A_initials));
    Optional<String> email = Optional.ofNullable(acct.getAttr(Provisioning.A_mail));
```

The behavior should be the same

Alternatively you can do the following:
```suggestion
    String sName = Optional.ofNullable(acct.getAttr(Provisioning.A_sn)).orElse("");
    String cName = Optional.ofNullable(acct.getAttr(Provisioning.A_cn)).orElse("");
    String initials = Optional.ofNullable(acct.getAttr(Provisioning.A_initials)).orElse("");
    String email = Optional.ofNullable(acct.getAttr(Provisioning.A_mail)).orElse("");
```

If you want the "fallback" value to be `""`

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
Polpetta: Yeah. Anyway as we discussed it is not possible to have it in the non-static block, so I'll resolve this point

Polpetta:
> updated the PR with all necessary changes

Yep, I answered to you comments, all the other stuff I marked it as resolved

store/src/java/com/zimbra/cs/account/ldap/LdapProvisioning.java
keshavbhatt: Offcourse.

store/src/java-test/com/zimbra/cs/account/ldap/LdapProvisioningTest.java
keshavbhatt: Will avoid them in the future.
Thanks

Polpetta (approved):
lgtm :+1:

PR #18, opened 2022-03-08
feat(carbonio-mailbox.hcl): add carbonio-preview (#18)

nomanAli95:
Since the previewer service has to be integrated with the mailbox in Carbonio, the carbonio-preview has been added to the carbonio-mailbox.hcl file to allow mailbox to communicate with preview via consul.

### DIFF
--- a/packages/appserver-service/carbonio-mailbox.hcl
+++ b/packages/appserver-service/carbonio-mailbox.hcl
@@ -12,6 +12,11 @@ services {
             destination_name   = "carbonio-storages"
             local_bind_port    = 20000
             local_bind_address = "127.78.0.7"
+          },
+          {
+            destination_name   = "carbonio-preview"
+            local_bind_port    = 20001
+            local_bind_address = "127.78.0.7"
           }
         ]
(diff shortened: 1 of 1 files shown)


M0Rf30 (approved):

packages/appserver-service/carbonio-mailbox.hcl
```diff
@@ -12,6 +12,11 @@ services {
             destination_name   = "carbonio-storages"
             local_bind_port    = 20000
             local_bind_address = "127.78.0.7"
+          },
+          {
+            destination_name   = "carbonio-preview"
```
keshavbhatt, line 17: I am not sure, but can you confirm the name of the destination is **carbonio-preview** and not **carbonio-previewer**?

nomanAli95: Yeah, I asked to the team which implemented the previewer service and they confirmed that the destination is **carbonio-preview**. The devguide doc has to be updated, we're already reported this.

PR #20, opened 2022-03-17
feat(SOAP): [CO-185] preview service SOAP endpoint

keshavbhatt:
- enables generate previews for mail attachment using Carbonio previewer service based on passed parameters.

### DIFF
--- a/store/src/java/com/zimbra/cs/service/mail/GetPreview.java
+++ b/store/src/java/com/zimbra/cs/service/mail/GetPreview.java
@@ -188,7 +188,6 @@ public class GetPreview extends MailDocumentHandler {
         }
       }
       post.setURI(URI.create(url));
-      log.info("K_PREVIEWER preview URL: " + url);
       MultipartEntityBuilder builder = MultipartEntityBuilder.create();
       builder.addBinaryBody("file", f, ContentType.create(contentType, "UTF-8"), f.getName());
       post.setEntity(builder.build());
@@ -312,7 +311,6 @@ public class GetPreview extends MailDocumentHandler {
       HttpClientBuilder clientBuilder = ZimbraHttpConnectionManager.getInternalHttpConnMgr()
           .newHttpClient();
       String url = baseURL + "/service/content/get?id=" + itemId + "&part=" + partNo + "&auth=co";
-      log.info("K_PREVIEWER using getAttachment URL: " + url);
       HttpGet get = new HttpGet(url);
       ZAuthToken zat = session.getAuthToken();
       Map<String, String> cookieMap = zat.cookieMap(false);
@@ -427,7 +425,6 @@ class ServiceResponse {
     try {
       File tempFile = File.createTempFile("preview_", "_" + origFileName);
       tempFilePath = tempFile.getAbsolutePath();
-      ZimbraLog.misc.info("K_PREVIEWER stored attachment in: " + tempFilePath);
       tempFile.deleteOnExit();
       FileOutputStream fos = new FileOutputStream(tempFile);
       fos.write(content);
--- a/store/src/java/com/zimbra/cs/service/mail/GetPreview.java
+++ b/store/src/java/com/zimbra/cs/service/mail/GetPreview.java
@@ -214,7 +214,7 @@ public class GetPreview extends MailDocumentHandler {
       String area = imageEle.getAttribute(MailConstants.A_P_AREA, ""); //mn
       String quality = imageEle.getAttribute(MailConstants.A_P_QUALITY, ""); //on
       String outputFormat = imageEle.getAttribute(MailConstants.A_P_OUTPUT_FORMAT, "jpeg"); //od
-      if (previewType.equals("thumbnail")) {
+      if ("thumbnail".equals(previewType)) {
         //add path arguments
         imageParams = "/" + area + "/" + previewType + "/";
         //add query params
--- a/store/src/java/com/zimbra/cs/service/mail/GetPreview.java
+++ b/store/src/java/com/zimbra/cs/service/mail/GetPreview.java
@@ -43,19 +43,23 @@ import org.apache.http.impl.client.HttpClientBuilder;
 import org.apache.http.impl.cookie.BasicClientCookie;
 import org.apache.http.util.EntityUtils;
 
+/**
+ * GetPreview class is registered to dispatch response to the requests coming from the
+ * GetPreviewRequest SOAP endpoint
+ */
 public class GetPreview extends MailDocumentHandler {
 
-  static final String PREVIEW_SERVICE_BASE_URL = "http://127.78.0.6:10000/";
+  private static final String PREVIEW_SERVICE_BASE_URL = "http://127.78.0.6:10000/";
   private final Log log = ZimbraLog.misc;
   private final String serverBaseUrl = getServerBaseUrl(null, false);
 
   /**
-   * Helper function to return the URL of server
+   * Helper method to return the baseURL for mailbox server
    *
    * @param server ({@link com.zimbra.cs.account.Server})
-   * @return url of server
+   * @return URL string of mailbox server
    */
-  public static String getServerBaseUrl(Server server, boolean withPort) {
+  private String getServerBaseUrl(Server server, boolean withPort) {
     String scheme = "https";
     try {
       server = server == null ? Provisioning.getInstance().getLocalServer() : server;
@@ -106,7 +110,7 @@ public class GetPreview extends MailDocumentHandler {
     }
 
     //Post it to the preview service with he passed parameters
-    if (attachmentResponse != null && attachmentResponse.statusCode == 200) {
+    if (attachmentResponse != null && attachmentResponse.getStatusCode() == 200) {
       try {
         previewResponse = getPreview(request, new File(attachmentResponse.getTempFilePath()));
         previewResponse.setOrigFileName(attachmentResponse.getOrigFileName());
@@ -124,8 +128,8 @@ public class GetPreview extends MailDocumentHandler {
     }
 
     //Pass the response in the soap response
-    if (previewResponse != null && previewResponse.statusCode == 200) {
-      String previewStream = Base64.getEncoder().encodeToString(previewResponse.content);
+    if (previewResponse != null && previewResponse.getStatusCode() == 200) {
+      String previewStream = Base64.getEncoder().encodeToString(previewResponse.getContent());
       Element previewDataStream = response.addUniqueElement("previewDataStream");
       previewDataStream.setText(previewStream);
       previewDataStream.addAttribute("file-name", previewResponse.getOrigFileName());
@@ -141,11 +145,12 @@ public class GetPreview extends MailDocumentHandler {
    *
    * @param f file to post to preview service(downloaded attachment in this case)
    * @return ServiceResponse ({@link ServiceResponse })
-   * @throws HttpException http exception
-   * @throws IOException   IO exception
+   * @throws HttpException HTTP exceptions occurred during the transport
+   * @throws IOException IO exception occurred during the transport
    */
-  private ServiceResponse getPreview(Element requestElement, File f) throws HttpException, IOException {
-    ServiceResponse serviceResponse = new ServiceResponse();
+  private ServiceResponse getPreview(Element requestElement, File f)
+      throws HttpException, IOException {
+    ServiceResponse previewServiceResponse = new ServiceResponse();
     HttpClientBuilder clientBuilder = ZimbraHttpConnectionManager.getInternalHttpConnMgr()
         .newHttpClient();
 
@@ -170,7 +175,8 @@ public class GetPreview extends MailDocumentHandler {
       } else {
         if (contentType.equalsIgnoreCase("image/png") || contentType.equalsIgnoreCase(
             "image/jpeg")) {
-          url = GetPreview.PREVIEW_SERVICE_BASE_URL + "preview/image/" + getImageParams(requestElement);
+          url = GetPreview.PREVIEW_SERVICE_BASE_URL + "preview/image/" + getImageParams(
+              requestElement);
         } else {
           url = GetPreview.PREVIEW_SERVICE_BASE_URL + "preview/pdf/" + getPdfParams(requestElement);
         }
@@ -181,14 +187,19 @@ public class GetPreview extends MailDocumentHandler {
       builder.addBinaryBody("file", f, ContentType.create(contentType, "UTF-8"), f.getName());
       post.setEntity(builder.build());
       HttpResponse response = HttpClientUtil.executeMethod(client, post);
-      serviceResponse.statusCode = response.getStatusLine().getStatusCode();
-      serviceResponse.content = EntityUtils.toByteArray(response.getEntity());
+      previewServiceResponse.setStatusCode(response.getStatusLine().getStatusCode());
+      previewServiceResponse.setContent(EntityUtils.toByteArray(response.getEntity()));
     } finally {
       post.releaseConnection();
     }
-    return serviceResponse;
+    return previewServiceResponse;
   }
 
+  /**
+   * Helper method to form request parameters for request type Image
+   * @param requestElement referenced IMage request element
+   * @return string containing request parameters
+   */
   private String getImageParams(Element requestElement) {
     String imageParams = "";
     Element imageEle = requestElement.getOptionalElement("image");
@@ -215,7 +226,11 @@ public class GetPreview extends MailDocumentHandler {
     return imageParams;
   }
 
-
+  /**
+   * Helper method to form request parameters for request type PDF
+   * @param requestElement referenced PDF request element
+   * @return string containing request parameters
+   */
   private String getPdfParams(Element requestElement) {
     String pdfParams = "";
     Element pdfEle = requestElement.getOptionalElement("pdf");
@@ -240,8 +255,12 @@ public class GetPreview extends MailDocumentHandler {
     return pdfParams;
   }
 
-  private Map<Integer, String> getPreviewServiceStatus() {
 
+  /**
+   * Method used to get the status(health/live) of PreviewService
+   * @return HashMap of containing StatusCode(int) and ReasonString(string)
+   */
+  private Map<Integer, String> getPreviewServiceStatus() {
     Map<Integer, String> status = new LinkedHashMap<>();
     HttpClientBuilder clientBuilder = ZimbraHttpConnectionManager.getInternalHttpConnMgr()
         .newHttpClient();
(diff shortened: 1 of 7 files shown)


store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,370 @@
...
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.config.SocketConfig;
+import org.apache.http.entity.ContentType;
+import org.apache.http.entity.mime.MultipartEntityBuilder;
+import org.apache.http.impl.client.BasicCookieStore;
+import org.apache.http.impl.client.HttpClientBuilder;
+import org.apache.http.impl.cookie.BasicClientCookie;
+import org.apache.http.util.EntityUtils;
+
+public class GetPreview extends MailDocumentHandler {
```
frisonisland, line 46: Hi Keshav, I would add the doc for this class to describe the general idea/usage of it.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,370 @@
...
+
+  static final String PREVIEW_SERVICE_BASE_URL = "http://127.78.0.6:10000/";
+  private final Log log = ZimbraLog.misc;
+  private final String serverBaseUrl = getServerBaseUrl(null, false);
+  Element requestElement = null;
```
frisonisland, line 51: I see we are storing the request as instance field, but I would avoid that because it's not clear what would happen on subsequent requests. 
If we take a look at the registration of this handler on the dispatcher it is initialized only one time.
So calling the handle() method multiple times could result in overwriting the request, potentially giving unwanted results.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,370 @@
...
+      HttpResponse response = HttpClientUtil.executeMethod(client, post);
+      serviceResponse.statusCode = response.getStatusLine().getStatusCode();
+      serviceResponse.content = EntityUtils.toByteArray(response.getEntity());
+    } finally {
+      post.releaseConnection();
+    }
+    return serviceResponse;
+  }
+
+  private String getImageParams() {
```
frisonisland, line 194: About this method, as per my comment above about request as field, I would change it to have it the request as parameter. Same can be said about other methods with same logic.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,370 @@
...
+    return serviceResponse;
+  }
+
+  private Element appendError(Element response, String errorMessage) {
+    Element error = response.addNonUniqueElement("error");
+    error.setText(errorMessage);
+    return response;
+  }
+
+  private static class ServiceResponse {
```
frisonisland, line 329: Also don't know if there is a specific need to make this class private, but I would move it to public and in a standalone file, with javadoc for everything (except setters and getters of course :) )

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,370 @@
...
+      return status;
+    } catch (HttpException | IOException e) {
+      status.put(0, "Carbonio preview service is not reachable.");
+    } finally {
+      get.releaseConnection();
+    }
+    return status;
+  }
+
+  private ServiceResponse getAttachment(String itemId, String partNo, String baseURL,
```
frisonisland, line 266: About JavaDoc, I would add it also for private methods.
Reason is in the future if we have to pickup the code again it will be easier to understand what we have to do!

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,370 @@
...
+  private final String serverBaseUrl = getServerBaseUrl(null, false);
+  Element requestElement = null;
+
+  /**
+   * Helper function to return the URL of server
+   *
+   * @param server ({@link com.zimbra.cs.account.Server})
+   * @return url of server
+   */
+  public static String getServerBaseUrl(Server server, boolean withPort) {
```
frisonisland, line 59: Not sure about this method and why we are using it with (null, false) parameters in field definition.
Also is it to get the url of the Files server/api?
I would either: 
* set serverbaseUrl in class constructor 
* set Server in constructor and maybe call something like serverInstance.getUrl() in the code.
Reason is testability, when we will be able to do it properly in the future this will be easier to do and maintain

frisonisland (changes requested):

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: Yeah sure.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: > So calling the handle() method multiple times could result in overwriting the request

The handle method is called by the dispatcher when a request arrives, we don't have to call it.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: This is to get the server's base URL where the mailbox is hosted. the first param is a server which if passed as null, we get the local one using provisioning. also the false is to set if we want the baseUrl with port or not. I have slightly modified this for my use case. Similar method signatures can be found in other places where we are trying to get the base URL of the local server.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
frisonisland: Yes, but it will call the handle() method of the same instance you provided with new()

keshavbhatt: It was just a generic XML Element variable, but I removed it now we passing it as the method's argument.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: Changes are in place, thanks :)

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: Docs are coming soon.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: changes are in place. Thanks

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,368 @@
...
+
+  public String getTempFilePath() {
+    return tempFilePath;
+  }
+
+  public byte[] getContent() {
+    return content;
+  }
+
+  public boolean storeContent(String filename, byte[] content) {
```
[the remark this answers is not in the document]
frisonisland: Ok, but keep in mind that when and if we have to use an SDK to call the preview API, the ServiceResponse object will be just a representation of the response, and it won't have logic about storing file, etc.
Having the logic on the GetPreview class it would be allow more decoupling and would be easier to refactor. Just a thought! 
Anyway open the PR when you feel it's ready, other people could provide other/different reasoning, so it's all good :)

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,441 @@
...
+              .addAttribute("status-message", entry.getValue());
+          previewServiceStatusCode.set(entry.getKey());
+        }
+    );
+
+    if (serverBaseUrl.isEmpty() || previewServiceStatusCode.intValue() != 200) {
+      return appendError(response, "Unable to get preview. Not able to get the server URL.");
+    }
+
+    String itemId = request.getAttribute("itemId", null);
```
frisonisland, line 103: There is a utility method that converts from Element to Class type for request and response object.
I would consider using that for type and compile time checking, so if something changes we know what to change.
It's zc.elementToJaxb method (zc is instance of ZimbraSoapContext in this scope).
To give you an idea: zc.elementToJaxb(request) -> jaxbElement.getItem()

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,441 @@
...
+    return previewServiceResponse;
+  }
+
+  /**
+   * Helper method to form request parameters for request type Image
+   *
+   * @param requestElement referenced IMage request element
+   * @return string containing request parameters
+   */
+  private String getImageParams(Element requestElement) {
```
frisonisland, line 209: Also here, if we use the ElementToJaxb, remember to reactor it to use GetPreviewRequest.
Also for other methods would be the same (getPdfParams, etc.)

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,441 @@
...
+    return imageParams;
+  }
+
+  /**
+   * Helper method to form request parameters for request type PDF
+   *
+   * @param requestElement referenced PDF request element
+   * @return string containing request parameters
+   */
+  private String getPdfParams(Element requestElement) {
```
frisonisland, line 241: So in these methods we are parsing the request and returning a query string.
I would consider changing the method to name to something like getPdfParamsAsQueryString or something like that.
Reason is it gives a better idea of what we are doing. 
If we don't want to be this specific (imagine the api changes in the future and we have to provide a JSON string instead of query), we can always provide the description in the JavaDoc.

frisonisland (changes requested):

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: using the utility class.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: changed the name if method,

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: we can pass the plain request element below our exception check, otherwise, we will also need SoapContext as a method param. If the request element will be malformed the exception will prevent further execution of this method.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,433 @@
...
+
+  /**
+   * Helper method to form request parameters for request type Image
+   *
+   * @param requestElement referenced IMage request element
+   * @return string containing request parameters
+   */
+  private String getImageParamsAsQueryString(Element requestElement) {
+    String imageParams = "";
+    Element imageEle = requestElement.getOptionalElement(MailConstants.E_P_IMAGE);
```
Polpetta, line 217: Are you sure this is the most elegant and maintainable way to access elements? I saw in @frisonisland codebase that it seems to be a method called `zsc.elementToJaxb(...)` to get the right request object instance.

This seems that this brings type safety inside this codebase which is always a good thing.

Polpetta (changes requested):
I've made some comments. I'll check it again once you add the tests too

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,433 @@
...
+
+    HttpClient client = clientBuilder.build();
+    HttpPost post = new HttpPost();
+    try {
+      String contentType = URLConnection.getFileNameMap().getContentTypeFor(f.getName());
+      String url;
+      if (contentType == null || contentType.isEmpty()) {
+        contentType = "application/pdf";
+        url = GetPreview.PREVIEW_SERVICE_BASE_URL + "preview/pdf/" + getPdfParamsAsQueryString(
+            requestElement);
```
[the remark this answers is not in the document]
keshavbhatt: Not sure, but we only support pdf and image. In this case, we are letting the preview service decide it, We also have error handling checks if the preview service refuse to generate preview.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,433 @@
...
+   */
+  private String getImageParamsAsQueryString(Element requestElement) {
+    String imageParams = "";
+    Element imageEle = requestElement.getOptionalElement(MailConstants.E_P_IMAGE);
+    if (imageEle != null) {
+      String previewType = imageEle.getAttribute(MailConstants.A_P_PREVIEW_TYPE, ""); //mn
+      String area = imageEle.getAttribute(MailConstants.A_P_AREA, ""); //mn
+      String quality = imageEle.getAttribute(MailConstants.A_P_QUALITY, ""); //on
+      String outputFormat = imageEle.getAttribute(MailConstants.A_P_OUTPUT_FORMAT, "jpeg"); //od
+      if (previewType.equals("thumbnail")) {
```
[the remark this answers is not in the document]
keshavbhatt: > This equals should be inverted, if previewType is `null` you get a `NullPointerException`

PreviewType cannot be null, we set it to empty string above. but I will keep the suggestion. Thanks

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,433 @@
...
+      tempFilePath = tempFile.getAbsolutePath();
+      tempFile.deleteOnExit();
+      FileOutputStream fos = new FileOutputStream(tempFile);
+      fos.write(content);
+      fos.close();
+      stored = true;
+    } catch (IOException ie) {
+      stored = false;
+    }
+    return stored;
```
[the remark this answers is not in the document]
keshavbhatt: It's no longer a POJO if we add logic to it, hence it's not a POJO :P 
It's a helper class that is being used in the handle method above.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,433 @@
...
+        }
+      } else {
+        throw ServiceException.RESOURCE_UNREACHABLE(
+            "Failed to get the attachment. Attachment provider service returned with code "
+                + attachmentResponse.getStatusCode(), null);
+      }
+    } catch (HttpException | IOException e) {
+      throw ServiceException.RESOURCE_UNREACHABLE(
+          "An exception occurred while completing your request.", e);
+    }
```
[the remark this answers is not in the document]
keshavbhatt: This is the best way and suits our use-case, Service exception is a valid exception here. I don't see any issue.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
keshavbhatt: We already used that in the handle method, if the element was not valid it would have returned an exception there.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,433 @@
...
+            + outputFormat;
+      } else {
+        String firstPage = pdfEle.getAttribute(MailConstants.A_P_FIRST_PAGE, "1"); //od
+        String lastPage = pdfEle.getAttribute(MailConstants.A_P_LAST_PAGE, "1"); //od
+        //add query params
+        pdfParams = "?first_page=" + firstPage + "&last_page=" + lastPage;
+      }
+    }
+    return pdfParams;
+  }
```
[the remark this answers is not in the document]
keshavbhatt: I responded there :P

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
```diff
@@ -0,0 +1,433 @@
...
+  /**
+   * stores content in a temp file.
+   *
+   * @param fileName filename received in content-disposition header whose data is being stored
+   * @param bytes    data in form of byte array
+   * @return true if content was stored correctly
+   */
+  public boolean storeContent(String fileName, byte[] bytes) {
+    boolean stored;
+    origFileName = fileName;
```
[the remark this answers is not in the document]
keshavbhatt: I don't get it. This class has only one responsibility -> manage and provide values returned in the response of services.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
Polpetta: Ok. The fact is regarding the `getOptionalElement(//ID)` stuff, that we wouldn't have if we converted the object to a request type in the first place. I dunno if I'm able to explain myself. If you want we can chat via call whenever you can :blush:

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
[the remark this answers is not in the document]
Polpetta: I mean you are doing side effect + doing something else. If you can I'd avoid to change the state of a class while calling a method, if said in a explicit way

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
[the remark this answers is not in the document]
Polpetta: You're right, but usually `Response` and `Request` kind of objects are POJOs, other operations on them are handled by other classes. In this way you're coupling a data structure with business logic, making more difficult in the future to maintain/work on them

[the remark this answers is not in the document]
keshavbhatt: My man, the first reason why we have this class is to have our object feature which can store the content in a temp location, take it as an intermediator which is adding some extra features to the response we got from the service. Otherwise, I would have used the HTTP response directly. Why I would wrap it into another POJO? I can explain the use case more clearly if you want in a quick call.

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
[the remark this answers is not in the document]
Polpetta: Solved by irl chat

store/src/java/com/zimbra/cs/service/mail/GetPreview.java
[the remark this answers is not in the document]
keshavbhatt: It will not cause any issues, we are just setting the filename, this response object is only modified once. Also, I cannot set the filename without getting the original response hence we cannot set the file in the class constructor while initializing object.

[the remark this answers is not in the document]
frisonisland: Just an update: I agree with @Polpetta on this. Also on an old change request I asked something very similar.
But as I see this class does not only provide values returned by services, but its also doing something on the filesystem hence it's mixing some logic with the data. 
Suggestion: I would have defined the storeContent method in the actual service class, which is GetPreview, and avoided side-effects.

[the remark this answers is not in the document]
keshavbhatt: I am not getting what you are guys talking about? Can we have a call so I can better understand your concerns?
Is the name of the class and what it is doing bothering us here? I am sure if we have named it something else then we would've not discussing this :D

Polpetta (approved):
LGTM :+1:

PR #28, opened 2022-04-11
Feat/copy to files api

frisonisland:
The goal of this PR is to provide the copy-to-files API to allow a user to upload an email attachment to Files.

**What's new: CopyToFiles** handler class + **Request and Response classes** + **registration in MailService** (store), MailConstants (common) and JaxbUtil (soap).

**Implementation**: 

- control flow **in functional programming style** using **vavr**, since it has been introduced from third-party carbonio-files-sdk. 
- **MailboxAttachmentService** class has been introduced to get an mail attachment, **instead of making a direct Http call to ContentServlet to get an attachment**.
- <s>we are loading the attachment in memory because Files SDK needs the attachment size and the only way to get it is to actually read the InputStream (open for other suggestion because I don't really like this)</s>. Now attachment size is calculated from stream and stream content is requested again.

**Testing:**

- **CopyToFilesIT** to test integration with Files SDK and soap exceptions.
- **CopyToFilesTest** for UT
- **Manual testing** (a lot), like **trying to get an attachment of another user, resulting in error** (in the current state was not able to automate this kind of tests). Same for authorization.

**Dependencies:** 

- **jackson** libraries updated to **2.13.1 in store module** (due to Files SDK conflict)
- **apache http client** updated **to 4.5.13** due to a runtime WebClassLoaderException (due to Files SDK conflict)

**Other changes**: 

- **many test classes have been updated** because Powermockito has been removed due to OpenJDK11 conflicts**. 
- **I created some methods in old classes** to return providers of instances, so the test can work without PowerMock and by mocking, **instead of calling static methods directly**. For example:  GalInfoProvider.getInstance -> getGalInforProvider()  new SoapHttpTransport() -> getSoapHttpTransportInstance()


soap/src/java/com/zimbra/soap/mail/message/CopyToFilesRequest.java
```diff
@@ -0,0 +1,80 @@
...
+
+@XmlAccessorType(XmlAccessType.NONE)
+@XmlRootElement(name= MailConstants.E_COPY_TO_FILES_REQUEST)
+public class CopyToFilesRequest {
+
+  /**
+   * @zm-api-field-tag messageId
+   * @zm-api-field-description ID of the email
+   */
+  @XmlElement(name=MailConstants.A_MESSAGE_ID /* messageId */, required=true)
```
keshavbhatt, line 22: Not a big deal for me personally but the `A` prefix is being used to indicate if the object is an Attribute in MailConstants.A_MESSAGE_ID while, I see we are trying to create an element that should be prefixed with `E` char.
Not a big deal, but just to stick with the convention we are following throughout the code.

store/src/java/com/zimbra/cs/service/MailboxAttachmentService.java
```diff
@@ -0,0 +1,36 @@
...
+ * Mailbox attachment provider
+ */
+public class MailboxAttachmentService implements AttachmentService {
+
+  public Try<MimePart> getAttachment(String accountId, AuthToken token, int messageId,
+      String part) {
+    return Try.of(() -> MailboxManager.getInstance().getMailboxByAccountId(accountId))
+        .mapTry(mailbox -> mailbox.getMessageById(new OperationContext(token), messageId))
+        .mapTry(message -> Mime.getMimePart(message.getMimeMessage(),
+            part))// have to read content to load the stream
```
keshavbhatt, line 22: comment could be dropped?

soap/src/java/com/zimbra/soap/mail/message/CopyToFilesRequest.java
frisonisland: I think you are right! 
Some commits ago I was using attributes on the request, but very recently had to change the implementation and left "A_" as naming without thinking about the meaning. Thanks

soap/src/java/com/zimbra/soap/mail/message/CopyToFilesRequest.java
```diff
@@ -0,0 +1,80 @@
+// SPDX-FileCopyrightText: 2022 Synacor, Inc.
```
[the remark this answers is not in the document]
frisonisland: I think this is a copy/paste mistake. Will check also the other classes. Thanks ;)

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -0,0 +1,171 @@
...
+    final ZimbraSoapContext zsc = getZimbraSoapContext(context);
+
+    Try<CopyToFilesRequest> copyToFilesRequestTry = getRequestObject(request);
+    Try<MimePart> mimePartTry = copyToFilesRequestTry.flatMap(
+        copyToFilesRequest -> getAttachmentToCopy(copyToFilesRequest, zsc));
+    String nodeId = Optional.ofNullable(
+            API.For(mimePartTry, copyToFilesRequestTry, Try.success(zsc))
+                .yield((mimePart, req, ctx) -> copyToFiles(mimePart, req, ctx).get())
+                .mapFailure(Case($(ex -> !(ex instanceof SoapFaultException)),
+                    new SoapFaultException("Service failure.", "", true)))
```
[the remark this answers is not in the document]
frisonisland: So the deal with SoapFaultException is that if the instance is created in specific ways it will return a 500 error client-side with an error message.
For example using the constructor new SoapFaultException("Service failure.", request) will produce a 200.
Unfortunately I'm not able to automate tests in a way that can be shown in the code.

Also I'm not a fan of defining inline SoapFaultException, and would have defined specific classes with error codes.
As you can see the SoapFaultException being used here has these parameters: message, error code, isServerFault
but error code is not defined. I will try to add more error codes if that's the case, else any suggestion is highly appreciated. The error codes should be meaningful for the client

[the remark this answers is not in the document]
keshavbhatt: I use ServiceException, it is able to handle all my needs :)
ex: ServiceException.FAILURE
      ServiceException INVALID_REQUEST etc check out there are lots of generic service failure methods with reasonable  
attributes you can pass into them.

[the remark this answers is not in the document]
keshavbhatt: the bonus is: Error code will be handled automatically if we use the correct method based on our reason.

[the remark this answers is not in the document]
frisonisland: Ok thanks, will check and update the code!

soap/src/java/com/zimbra/soap/mail/message/CopyToFilesRequest.java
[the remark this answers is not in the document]
frisonisland: Mark as resolved because I have removed the Synacor line (my bad)

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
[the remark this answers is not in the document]
frisonisland: Ok, I have used a mix of ServiceException.FAILURE, PARSE_ERROR and NOT_FOUND.
Also introduced a static method on ServiceException to allow calling constructor without throwable parameter.

soap/src/java/com/zimbra/soap/mail/message/CopyToFilesRequest.java
frisonisland: This has been updated: my idea is to leave identifiers as attributes (like the id), everything else as element.
In this case mailId and Part uniquely identify a CopyToFilesRequest, so they are attributes. Let me know

Polpetta:
Please add the following `pre-commit` script checker, so that we can start checking what we're writing:
```yaml
# See https://pre-commit.com for more information
# See https://pre-commit.com/hooks.html for more hooks
repos:
-   repo: https://github.com/pre-commit/pre-commit-hooks
    rev: v4.1.0
    hooks:
      - id: trailing-whitespace
      - id: end-of-file-fixer
      - id: check-yaml
      - id: check-added-large-files
      - id: check-toml
      - id: mixed-line-ending
- repo: https://github.com/gruntwork-io/pre-commit
  rev: v0.1.17
  hooks:
      - id: shellcheck
- repo: https://github.com/macisamuele/language-formatters-pre-commit-hooks
  rev: v2.3.0
  hooks:
      - id: pretty-format-java
        args: [--autofix]
- repo: https://github.com/igorshubovych/markdownlint-cli
  rev: v0.31.1
  hooks:
    - id: markdownlint
    - id: markdownlint-fix
- repo: https://github.com/jorisroovers/gitlint
  rev:  v0.17.0
  hooks:
    - id: gitlint
```

Remember to install it with `pre-commit install`. You can then squash your commit (if you want) and it will check only on your diff

store/src/java-test/com/zimbra/cs/service/mail/CopyToFilesIT.java
```diff
@@ -0,0 +1,321 @@
...
+    up.setPart("2");
+    Element element = JaxbUtil.jaxbToElement(up);
+
+    try {
+      copyToFiles.handle(element, context);
+      fail("Did not throw SoapFault exception.");
+    } catch (ServiceException serviceException) {
+      assertEquals("File not found.", serviceException.getMessage());
+    }
+  }
```
Polpetta, line 144: This is not the way to check for exceptions in the codebase: please check https://www.baeldung.com/junit-assert-exception

In particular note that with:
```java
@Rule
public ExpectedException exceptionRule = ExpectedException.none();
```

After you can check the exception message and other stuff

store/src/java-test/com/zimbra/cs/service/mail/CopyToFilesIT.java
```diff
@@ -0,0 +1,321 @@
...
+    up.setMessageId("1");
+    up.setPart("2");
+    Element element = JaxbUtil.jaxbToElement(up);
+
+    try {
+      copyToFiles.handle(element, context);
+      fail("Did not throw SoapFault exception.");
+    } catch (ServiceException serviceException) {
+      assertEquals("system failure: internal error.", serviceException.getMessage());
+    }
```
Polpetta, line 190: This is not the way to check for exceptions in the codebase: please check https://www.baeldung.com/junit-assert-exception

In particular note that with:
```java
@Rule
public ExpectedException exceptionRule = ExpectedException.none();
```

After you can check the exception message and other stuff

store/src/java-test/com/zimbra/cs/service/mail/CopyToFilesIT.java
```diff
@@ -0,0 +1,321 @@
...
+    up.setMessageId("123");
+    up.setPart("Whatever you want");
+    Element element = JaxbUtil.jaxbToElement(up);
+
+    try {
+      copyToFiles.handle(element, context);
+      fail("Did not throw SoapFault exception.");
+    } catch (ServiceException serviceException) {
+      assertEquals(
+          "system failure: got null response from Files server.", serviceException.getMessage());
```
Polpetta, line 237: This is not the way to check for exceptions in the codebase: please check https://www.baeldung.com/junit-assert-exception

In particular note that with:
```java
@Rule
public ExpectedException exceptionRule = ExpectedException.none();
```

After you can check the exception message and other stuff

store/src/java-test/com/zimbra/cs/service/mail/CopyToFilesIT.java
```diff
@@ -0,0 +1,321 @@
...
+    }
+
+    verify(mockFilesClient, Mockito.times(1))
+        .uploadFile(
+            ZimbraCookie.COOKIE_ZM_AUTH_TOKEN + "=" + authToken.getEncoded(),
+            "FOLDER_1",
+            fileName,
+            contentType,
+            uploadContent,
+            fileSize);
```
Polpetta, line 297: This is not the way to check for exceptions in the codebase: please check https://www.baeldung.com/junit-assert-exception

In particular note that with:
```java
@Rule
public ExpectedException exceptionRule = ExpectedException.none();
```

After you can check the exception message and other stuff

store/src/java-test/com/zimbra/cs/service/mail/ModifyProfileImageTest.java
```diff
@@ -109,13 +103,13 @@ public void testZCS3545() throws Exception {
...
+//            .toReturn("image/jpeg");
         byte[] bytes = IOUtils.toByteArray(up.getInputStream());
-        PowerMockito
-            .stub(PowerMockito.method(NativeFormatter.class, "getResizedImageData",
-                InputStream.class, String.class, String.class, Integer.class, Integer.class))
-            .toReturn(bytes);
+//        Mockito
+//            .stub(Mockito.method(NativeFormatter.class, "getResizedImageData",
+//                InputStream.class, String.class, String.class, Integer.class, Integer.class))
+//            .toReturn(bytes);
```
Polpetta, line 112: Remove commented code if not necessary

store/src/java/com/zimbra/cs/mailbox/MailServiceException.java
```diff
@@ -92,6 +92,8 @@ public class MailServiceException extends ServiceException {
 
     public static final String INVALID_COMMIT_ID = "mail.INVALID_COMMIT_ID";
     public static final String INVALID_IMAGE = "mail.INVALID_IMAGE";
+    //TODO: in future we can consider moving not implemented to a more generic/shared error code
```
Polpetta, line 95: This TODO is still valid?

store/src/java/com/zimbra/cs/service/AttachmentService.java
```diff
@@ -0,0 +1,24 @@
+package com.zimbra.cs.service;
+
+import com.zimbra.cs.account.AuthToken;
+import io.vavr.control.Try;
+import java.io.InputStream;
+import javax.mail.internet.MimePart;
+
+/**
+ * Interface for attachment content provider (get).
+ * @author davidefrison
```
Polpetta, line 10: Add `@since` specification

store/src/java/com/zimbra/cs/service/MailboxAttachmentService.java
```diff
@@ -0,0 +1,23 @@
...
+import com.zimbra.cs.account.AuthToken;
+import com.zimbra.cs.mailbox.MailboxManager;
+import com.zimbra.cs.mailbox.OperationContext;
+import com.zimbra.cs.mime.Mime;
+import io.vavr.control.Try;
+import javax.mail.internet.MimePart;
+
+/**
+ * Mailbox attachment provider
+ * @author davidefrison
```
Polpetta, line 12: Add `@since` part

Polpetta (changes requested):
I've still to review `CopyToFiles` class, but right now there are those changes to do :grin:

store/src/java-test/com/zimbra/cs/service/mail/CopyToFilesIT.java
frisonisland: Ok thanks, was looking for assertThrows but forgot to check what was the JUnit 4 way.

store/src/java/com/zimbra/cs/mailbox/MailServiceException.java
frisonisland: I'd say yes. My idea is simply having a "not implemented" for generic purposes (e.g.: we want to provide a new API for some early access to the client but the functionality is not quite there).

store/src/java/com/zimbra/cs/service/MailboxAttachmentService.java
frisonisland: Added. But we should decide/talk about versioning ;)

soap/src/java/com/zimbra/soap/mail/message/CopyToFilesRequest.java
```diff
@@ -0,0 +1,87 @@
...
+import javax.xml.bind.annotation.XmlAttribute;
+import javax.xml.bind.annotation.XmlElement;
+import javax.xml.bind.annotation.XmlRootElement;
+
+/**
+ * Class that identifies a CopyToFiles request.
+ * The idea is that message id and part are identifiers, so they are attributes.
+ * File description and folder id are elements of the request.
+ * @author davidefrison
+ *
```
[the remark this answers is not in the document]
frisonisland: Added

store/src/java-test/com/zimbra/cs/filter/ZimbraMailAdapterTest.java
```diff
@@ -36,7 +31,7 @@ public static void init() throws Exception {
     public void testUpdateIncomingBlob() throws Exception{
         int mboxId = 10;
 
-        Mailbox mbox = PowerMockito.mock(Mailbox.class);
+        Mailbox mbox = Mockito.mock(Mailbox.class);
```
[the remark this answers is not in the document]
frisonisland: Done

store/src/java/com/zimbra/cs/service/AttachmentService.java
frisonisland: Done

store/src/java/com/zimbra/cs/service/MailboxAttachmentService.java
Polpetta: Yeah I agree with you

store/src/java-test/com/zimbra/cs/ldap/unboundid/UBIDUserCertificateAttributeTest.java
```diff
@@ -4,44 +4,49 @@
...
+    Entry entry = mock(Entry.class);
+    UBIDAttributes attributes = new UBIDAttributes(entry);
+    when(entry.getAttribute(ContactConstants.A_userCertificate))
+        .thenReturn(null); // entry does not contain "userCertificate" attribute
+    when(entry.getAttribute(lookupAttr))
+        .thenReturn(attr); // entry contains "userCertificate;binary" attribute
+    try {
+      assertEquals(attributes.getMultiAttrString(lookupAttr, false)[0], certBase64);
+    } catch (com.zimbra.cs.ldap.LdapException e) {
+      fail("Exception thrown");
```
Polpetta, line 33: I mean this is not really necessary, the test will fail even if the exception is thrown, so this is sort-of unnecessary code

store/src/java-test/com/zimbra/cs/ldap/unboundid/UBIDUserCertificateAttributeTest.java
```diff
@@ -4,44 +4,49 @@
...
+    Entry entry = mock(Entry.class);
+    UBIDAttributes attributes = new UBIDAttributes(entry);
+    when(entry.getAttribute(lookupAttr))
+        .thenReturn(null); // entry does not contain "userCertificate;binary" attribute
+    when(entry.getAttribute(ContactConstants.A_userCertificate))
+        .thenReturn(attr); // entry contains "userCertificate" attribute
+    try {
+      assertEquals(attributes.getMultiAttrString(lookupAttr, false)[0], certBase64);
+    } catch (com.zimbra.cs.ldap.LdapException e) {
+      fail("Exception thrown");
```
Polpetta, line 49: Same as before here

store/src/java-test/com/zimbra/cs/service/mail/ModifyProfileImageTest.java
```diff
@@ -4,184 +4,163 @@
...
+    Assert.assertEquals(ldapThumbnailImageData, attrMap.get(ZAttrProvisioning.A_thumbnailPhoto));
```
Polpetta, line 114: Static import

store/src/java-test/com/zimbra/cs/service/mail/ModifyProfileImageTest.java
```diff
@@ -4,184 +4,163 @@
...
+    Element response =
+        new ModifyProfileImage().handle(request, ServiceTestUtil.getRequestContext(acct));
+    ModifyProfileImageResponse responseJaxb = JaxbUtil.elementToJaxb(response);
+    int profileItemId = responseJaxb.getItemId();
+    GetInfoRequest getRequestJaxb = new GetInfoRequest();
+    getRequestJaxb.setSections("mbox");
+    Element getRequest = JaxbUtil.jaxbToElement(getRequestJaxb);
+    Element getResponse = new GetInfo().handle(getRequest, ServiceTestUtil.getRequestContext(acct));
+    GetInfoResponse getResponseJaxb = JaxbUtil.elementToJaxb(getResponse);
+    Assert.assertEquals(profileItemId, getResponseJaxb.getProfileImageId());
```
Polpetta, line 137: Static import

store/src/java-test/com/zimbra/cs/service/mail/ModifyProfileImageTest.java
```diff
@@ -4,184 +4,163 @@
...
+    new ModifyProfileImage().handle(request1, ServiceTestUtil.getRequestContext(acct));
+    GetInfoRequest getRequestJaxb1 = new GetInfoRequest();
+    getRequestJaxb1.setSections("mbox");
+    Element getRequest1 = JaxbUtil.jaxbToElement(getRequestJaxb1);
+    Element getResponse1 =
+        new GetInfo().handle(getRequest1, ServiceTestUtil.getRequestContext(acct));
+    // check that the profile image information is not present in GetInfo as well as LDAP
+    Assert.assertEquals(false, getResponse1.prettyPrint().contains("profileImageId"));
+    Map<String, Object> attrMap1 = acct.getAttrs(false, false);
+    Assert.assertEquals(null, attrMap1.get(ZAttrProvisioning.A_thumbnailPhoto));
```
Polpetta, line 155: Static imports

store/src/java-test/com/zimbra/cs/service/mail/ModifyProfileImageTest.java
```diff
@@ -4,184 +4,163 @@
...
-
-    @After
-    public void tearDown() {
-        try {
-            MailboxTestUtil.clearData();
-        } catch (Exception e) {
-            e.printStackTrace();
-        }
-    }
+  }
```
Polpetta, line 165: Don't catch exceptions on `@Before`, `@After` or `@BeforeClass` and similar. Let the exception be sent to JUnit, otherwise possible errors in tests could be not displayed because they're catched

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -55,120 +54,183 @@ public CopyToFiles(
...
+                                fileName,
+                                contentType,
+                                attachmentStream,
+                                attachmentSize))
+                    .flatMap(x -> x)
+                    .onFailure(ex -> mLog.debug(ex.getMessage()))
+                    .mapFailure(
+                        Case(
+                            $(ex -> !(ex instanceof ServiceException)),
+                            ex -> ServiceException.FAILURE("internal error.", ex)))
```
Polpetta, line 108: You can use `onFailure` method to avoid `instanceOf` that is not for being slow.
```suggestion
                    .onFailure(ServiceException.class, ex -> ServiceException.FAILURE("internal error.", ex))
```

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -55,120 +54,183 @@ public CopyToFiles(
...
    */
   private Try<CopyToFilesRequest> getRequestObject(Element request) {
     return Try.<CopyToFilesRequest>of(() -> JaxbUtil.elementToJaxb(request))
         .onFailure(ex -> mLog.debug(ex.getMessage()))
-        .mapFailure(Case($(instanceOf(Exception.class)),
-            ex -> ServiceException.PARSE_ERROR("Malformed request.", ex)));
+        .mapFailure(
+            Case(
+                $(instanceOf(Exception.class)),
+                ex -> ServiceException.PARSE_ERROR("Malformed request.", ex)));
```
Polpetta, line 130: See suggestion above:
```suggestion
        .onFailure(Exception.class, ex -> ServiceException.PARSE_ERROR("Malformed request.", ex));
```

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -55,120 +54,183 @@ public CopyToFiles(
...
+        Try.of(() -> Integer.parseInt(request.getMessageId()))
             .onFailure(ex -> mLog.debug(ex.getMessage()))
-            .mapFailure(Case($(instanceOf(Exception.class)),
-                ex -> ServiceException.NOT_FOUND("File not found.", ex)))).get();
+            .mapFailure(
+                Case(
+                    $(instanceOf(Exception.class)),
+                    ex ->
+                        ServiceException.PARSE_ERROR(
+                            MailConstants.A_MESSAGE_ID + " must be an integer.", ex)));
```
Polpetta, line 150: See suggestion above:
```suggestion
            .onFailure(Exception.class,
                    ex ->
                        ServiceException.PARSE_ERROR(
                            MailConstants.A_MESSAGE_ID + " must be an integer.", ex));
```

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -55,120 +54,183 @@ public CopyToFiles(
...
+                    .getAttachment(
+                        context.getAuthtokenAccountId(),
+                        context.getAuthToken(),
+                        messageId,
+                        req.getPart())
+                    .onFailure(ex -> mLog.debug(ex.getMessage()))
+                    .mapFailure(
+                        Case(
+                            $(instanceOf(Exception.class)),
+                            ex -> ServiceException.NOT_FOUND("File not found.", ex))))
```
Polpetta, line 165: See suggestion above:
```suggestion
                    .onFailure(Exception.class,
                            ex -> ServiceException.NOT_FOUND("File not found.", ex))
```

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -55,120 +54,183 @@ public CopyToFiles(
...
-      return fileSize;
-    });
+    return Try.of(
+        () -> {
+          long fileSize = 0;
+          byte[] buffer = new byte[8192];
+          for (int read = inputStream.read(buffer); read != -1; read = inputStream.read(buffer)) {
+            fileSize += read;
+          }
+          return fileSize;
```
Polpetta, line 184: Suggestion: IIRC there is an Apache Commons method or a Stream way to avoid doing it the ol' imperative way ;)

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -55,120 +54,183 @@ public CopyToFiles(
...
+   * @param attachment {@link MimePart} attachment
+   * @return try of attachment content type
+   */
+  private Try<String> getAttachmentContentType(MimePart attachment) {
+    return Try.of(() -> attachment.getContentType())
+        .onFailure(ex -> mLog.debug(ex.getMessage()))
+        .mapFailure(
+            Case(
+                $(instanceOf(Exception.class)),
+                ex -> ServiceException.FAILURE("Cannot get file content-type.", ex)));
```
Polpetta, line 234:
```suggestion
        .onFailure(Exception.class,
                ex -> ServiceException.FAILURE("Cannot get file content-type.", ex));
```

store/src/java-test/com/zimbra/cs/service/UserServletTest.java
```diff
@@ -14,384 +18,372 @@
...
+      when(request.getRequestURI()).thenReturn("service/home/");
+      when(request.getParameter("auth")).thenReturn("basic");
+      when(request.getParameter("loc")).thenReturn("en_US");
+      when(request.getHeader("Authorization")).thenReturn("Basic dGVzdDM0ODg6dGVzdDEyMw==");
+      when(request.getQueryString()).thenReturn("auth=basic&view=text&id=261");
+
+      userServlet.doGet(request, response);
+      assertEquals(401, response.getStatus());
+      //            Commenting until we can figure out why this fails in CI env.
+      //            Assert.assertEquals("must authenticate", response.getMsg());
```
Polpetta, line 69: Question: have you figured out why? :sweat_smile: It is important to investigate it if it fails on the CI since it means tests runs in a different manner in CI or your machine

store/src/java-test/com/zimbra/cs/service/UserServletTest.java
```diff
@@ -14,384 +18,372 @@
...
+    } catch (Exception e) {
+      e.printStackTrace();
+      fail("No exception should be thrown.");
+    }
```
Polpetta, line 73: Please let the exception fly up in the junit sky :D

Polpetta (changes requested):
I reviewed your latest changes and there are some things to fix here and there.

Little note: it is ok to put documentation on a tests if it helps the next maintainer to understand it, but personally, I don't consider it mandatory. Good job writing it down anyway :+1:

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
frisonisland: @Polpetta I'm not sure this will introduce the behavior I was thinking about.
The idea is:

- to log the underlying exception
- map the exception to a specific ServiceException.

By using the onFailure(), which accepts a consumer, it will execute the consumer in case it matches but it will return the identity (this). So it will not map the exception to a ServiceException and the old failure will be preserved. 
Also some tests have been made to track this behavior and if I change the code as suggested they fail.(see shouldThrowServiceExceptionIfFileNotFound)

What I can do (but it's a style thing at this point) is:
```suggestion
        .mapFailure(
            Case(
                $(instanceOf(Exception.class)),
                ex -> {
                              mLog.debug(ex.getMessage());
                              return ServiceException.NOT_FOUND("File not found.", ex);
                            }));
```
and remove the onFailure entirely.

store/src/java-test/com/zimbra/cs/ldap/unboundid/UBIDUserCertificateAttributeTest.java
frisonisland: @Polpetta first of all: sorry for this monster PR. 
**The changes I made here are because I removed PowerMockito** and had to change many tests accordingly. **But these tests were not made by me.** This is one of those test classes. The only tests made by me are the one inside **CopyToFilesIT**.
We can talk about removing these tests if they do not provide any value, but I'll propose to save it for the next iteration, else there will be a never-ending refactor :)

Same can be said about static import in many places. But in that case there is no harm in adapting the code style to our needs, so I'll gladly do it.

frisonisland: Let me add: one could argue that these files should not be changed if there is no added value.

Value: we can run some old tests without them breaking because of PowerMockito. 
Also we are laying a base to slowly improve the test codebase.

store/src/java-test/com/zimbra/cs/service/UserServletTest.java
frisonisland: @Polpetta soory again, I think it's better to see it together, because things have got out of hand.
These changes are because I removed PowerMockito, but can't answer about everything else (e.g.: catching exceptions, commenting why it fails in CI, etc.). It's the first part of a refactor.
Also it's definitely not my style :sweat_smile:

store/src/java-test/com/zimbra/cs/ldap/unboundid/UBIDUserCertificateAttributeTest.java
Polpetta: Yeah, I agree with you, since this is a recent fork we should adapt it gradually, and in my opinion the best way to do it is to adapt what we're touching along the way, otherwise it'd impossible to do it at once. So no problem for the exception, if you want remove it otherwise it is fine (but someone else sooner or later will have to do it)

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
Polpetta: Ooooh ok, I got the point. Nope, keep it like it is and resolve all the other corrections pleasee :+1: Resolve this one once you've read the comment

store/src/java-test/com/zimbra/cs/service/mail/SearchActionTest.java
```diff
@@ -44,122 +32,133 @@
 import com.zimbra.soap.mail.message.SearchRequest;
 import com.zimbra.soap.mail.type.BulkAction;
 import com.zimbra.soap.type.SearchHit;
-
+import java.util.List;
```
frisonisland, line 35: **FYI (if you are here wondering what changed)**: this class has been updated due to pre-commit linting

store/src/java-test/com/zimbra/cs/imap/MemcachedImapCacheTest.java
```diff
@@ -7,9 +7,15 @@
 
 import static org.junit.Assert.assertNull;
 import static org.junit.Assert.fail;
+import static org.mockito.Mockito.mock;
```
frisonisland, line 10: **FYI:** class changed due to pre-commit

store/src/java-test/com/zimbra/cs/service/mail/CopyToFilesIT.java
```diff
@@ -0,0 +1,311 @@
+package com.zimbra.cs.service.mail;
```
frisonisland, line 1: **FYI:** class changed due to pre-commit

store/src/java/com/zimbra/cs/service/mail/CopyToFiles.java
frisonisland: Thanks, I checked the Apache library, but all I could find was methods that load the stream into a byte array and then can compute the size.
I did it this way because I don't need the entire byte array and it will avoid storing too much in memory.
Also not sure about other stream way, I think it will require to read, create a list, stream it, but the result will be the same, if not it will make the code less readable in this particular case.

store/src/java-test/com/zimbra/cs/mailbox/ContactTest.java
```diff
@@ -13,30 +13,6 @@
 import static org.mockito.Mockito.mock;
```
Polpetta, line 13: Pre commit here too?

Polpetta (approved):
LGTM :+1:

store/src/java-test/com/zimbra/cs/mailbox/ContactTest.java
frisonisland: Yes, It's the CopyToFIlesIT file that is new (the very next one), sorry for the mistake in the comment!

keshavbhatt (approved):

PR #30, opened 2022-04-14
Included template expansion for messaging configs

apontini:
Chats messaging configs for nginx were added to the templates list to allow for expansion
with the proxy restart


Polpetta (approved):
LGTM even if I don't understand very well what's happening

keshavbhatt (approved):

PR #33, opened 2022-04-29
Feat/CO-168 stats compatible with prometheus

keshavbhatt:
These changes Make mailbox stats compatible with Prometheus metrics.

### DIFF
--- a/common/src/java-test/com/zimbra/common/stats/PrometheusFormatterTest.java
+++ b/common/src/java-test/com/zimbra/common/stats/PrometheusFormatterTest.java
@@ -8,7 +8,7 @@ import org.junit.Test;
 public class PrometheusFormatterTest {
 
   @Test
-  public void ShouldFormatCorrectlyWhenCalledProcessComplexStat() {
+  public void shouldFormatCorrectlyWhenCalledProcessComplexStat() {
 
     StringBuilder buf = new StringBuilder();
     String[] header = {"command", "exec_count", "exec_ms_avg"};
@@ -24,7 +24,7 @@ public class PrometheusFormatterTest {
   }
 
   @Test
-  public void ShouldCleanHeaderWhenCalledSanitizeHeader() {
+  public void shouldCleanHeaderWhenCalledSanitizeHeader() {
     String[] headers = {"command", "exec-count", "exec_ms-avg", "exec's_duration"};
     StringBuilder buf = new StringBuilder();
     IntStream.range(1, headers.length)
@@ -34,7 +34,7 @@ public class PrometheusFormatterTest {
   }
 
   @Test
-  public void ShouldFormatCorrectlyWhenCalledProcessNormalStats() {
+  public void shouldFormatCorrectlyWhenCalledProcessNormalStats() {
     String[] headers = {
       "AnonymousIoService",
       "CloudRoutingReaderThread",
@@ -75,7 +75,7 @@ public class PrometheusFormatterTest {
   }
 
   @Test
-  public void AddTimestampWhenCalledAddTimeStamp() {
+  public void shouldAddTimestampWhenCalledAddTimeStamp() {
     final long timestamp = 1651670178388L;
     var logBuffer = new StringBuilder();
     PrometheusFormatter.addTimeStamp("sql", logBuffer, timestamp);
--- a/common/src/java-test/com/zimbra/common/stats/PrometheusFormatterTest.java
+++ b/common/src/java-test/com/zimbra/common/stats/PrometheusFormatterTest.java
@@ -1,0 +1,85 @@
+package com.zimbra.common.stats;
+
+import static org.junit.Assert.assertEquals;
+
+import java.util.stream.IntStream;
+import org.junit.Test;
+
+public class PrometheusFormatterTest {
+
+  @Test
+  public void ShouldFormatCorrectlyWhenCalledProcessComplexStat() {
+
+    StringBuilder buf = new StringBuilder();
+    String[] header = {"command", "exec_count", "exec_ms_avg"};
+    String[] stats = {"NoOpRequest", "2", "0"};
+
+    PrometheusFormatter.processComplexStats("soap", buf, header, stats);
+
+    String expected =
+        "soap_exec_count {command=\"NoOpRequest\"} 2\n"
+            + "soap_exec_ms_avg {command=\"NoOpRequest\"} 0\n";
+
+    assertEquals(expected, buf.toString());
+  }
+
+  @Test
+  public void ShouldCleanHeaderWhenCalledSanitizeHeader() {
+    String[] headers = {"command", "exec-count", "exec_ms-avg", "exec's_duration"};
+    StringBuilder buf = new StringBuilder();
+    IntStream.range(1, headers.length)
+        .forEach(i -> buf.append(PrometheusFormatter.sanitizeHeader(headers[i])).append(","));
+    String expected = "exec_count,exec_ms_avg,execs_duration,";
+    assertEquals(expected, buf.toString());
+  }
+
+  @Test
+  public void ShouldFormatCorrectlyWhenCalledProcessNormalStats() {
+    String[] headers = {
+      "AnonymousIoService",
+      "CloudRoutingReaderThread",
+      "GC",
+      "ImapSSLServer",
+      "ImapServer",
+      "LmtpServer",
+      "Pop3SSLServer",
+      "Pop3Server",
+      "ScheduledTask",
+      "SocketAcceptor",
+      "Thread",
+      "Timer",
+      "btpool",
+      "pool"
+    };
+
+    String[] stats = {"0", "0", "0", "0", "0", "2", "0", "0", "10", "0", "2", "5", "0", "0"};
+
+    StringBuilder buf = new StringBuilder();
+    PrometheusFormatter.processNormalStats("threads", buf, headers, stats);
+    String expected =
+        "threads_AnonymousIoService 0\n"
+            + "threads_CloudRoutingReaderThread 0\n"
+            + "threads_GC 0\n"
+            + "threads_ImapSSLServer 0\n"
+            + "threads_ImapServer 0\n"
+            + "threads_LmtpServer 2\n"
+            + "threads_Pop3SSLServer 0\n"
+            + "threads_Pop3Server 0\n"
+            + "threads_ScheduledTask 10\n"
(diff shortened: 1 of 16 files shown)


common/src/java/com/zimbra/common/stats/StatsDumper.java
```diff
@@ -13,153 +15,143 @@
...
+      if (mDataSource.hasTimestampColumn()) {
+        writer.write("timestamp,");
+      }
+      writer.write(header);
+      writer.write("\n");
+    }
+
+    // Write data and close
+    writer.write(buf.toString());
+    writer.close();
```
frisonisland, line 127: I would consider to close it on a finally. Maybe with a try resources.
This is because if any exception happens at least we can close the file.

common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java
```diff
@@ -0,0 +1,181 @@
...
+  }
+
+  /**
+   * Schedules a new stats task in a thread & calls the dumper to write collected data to the stats
+   * file
+   *
+   * @param dataSource the data source {@link StatsDumperDataSource}
+   * @param intervalMillis interval between writes. The first write is delayed by this interval.
+   */
+  public static void schedule(final StatsDumperDataSource dataSource, final long intervalMillis) {
```
frisonisland, line 34: Ok, I'm seeing that some logic of StatsDumperPrometheus is the same as StatsDumper.
Also seeing StatsDumper and StatsDumperPrometheus makes me think that one is an implementation of another (or a specific case).

At least I would consider a class extension since there are some common patterns: file directory exists, same runnable, etc.

common/src/java/com/zimbra/common/stats/StatsDumper.java
keshavbhatt: I have made the suggested changes. 
Note that we don't want to handle any exception in this case, if some exception occurs they are passed to the method.

common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java
keshavbhatt: Yes, I agree and it makes 100% sense. But I did not consider this while extracting logic from StatsDumper to implement something similar called StatsDumperPrometheus. I thought we might need to change the behavior at some point maybe in the future.
Also, I think tightly coupling a Schedule class will prevent us from changing the scheduling of two different dumpers who can have different requirements(that's why we have two of them separated). I hope I am clear, we can have a quick call about this later if we want.

frisonisland: I get your point.
But consider that having a simple Scheduler class will not introduce tight couple, but more flexibility instead.
My first (wrong) suggestion  of extending StatsDumper would have introduced tight coupling (inheritance), while having a Scheduler class + Callable will favor composition and flexibility. The two Callable WILL have different logic :)
For example this way we could:
* define a common Scheduler if we want to maintain the same clock cycle (schedule interval).
* define two different Scheduler if we want to have different intervals for Prometheus stats and standard stats
* have specific schedule interval for Imap, Pop, etc. for example

Something like
```
public class StatsScheduler {
        
        private final long interval;
        public StatsScheduler(long interval) {
            this.interval = interval;
        }
        
        public void schedule(Callable<Void> callable) {
            // your logic about logging, etc.
            Try.of(callable::call);
        }
    }```
    and we will recycle the code in the schedule() method. Just a suggestion.

Comment (keshavbhatt) [reply common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java:34]: You are right but if you take a look at the ZimbraPref class we are scheduling dumpers and not the stats data source. Stats data sources bring the stats data when different services push them. Then when the dumper runs at a defined interval they collect the data from different sources and write them to the destination. 

We can use what you are proposing but I don't see any major benefits. Also, note that going this way will affect various other implementations that already exist and that have nothing to do with this PR.

Comment (frisonisland) [reply common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java:34]: My idea was to schedule the dumpers, not the data sources! That's why I proposed schedule(Callable) -> the callable is the dumper.
The "only" benefits I see are: more flexibility + easier testable code (more unit). You could even assert that Prometheus dumper is indeed in the correct format, without starting a schedule or a thread.
Also I proposed these changes because I saw that in ZimbraPerf class we call StatsDumper.schedule, or StatsDumperProm.schedule.
Instead we could have statsScheduler.schedule(stdDumper), or statsScheduler.schedule(promDumper), where statsScheduler = new StatsScheduler(interval).
I assumed this was the place were its all tied together, so it didn't seem too much of a problem.
Anyway feel free to consider or reject the suggestion, not a blocker, but I think many developers will be grateful if we can favor composition and break things into smaller pieces (even if it will require a bit more effort)!

**Review by Polpetta — COMMENTED** [review]
A part from the scheduler stuff the rest is fine. I've commented some tip and trick here and there

File: common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java (hunk 2 above)
Comment (keshavbhatt) [reply common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java:34]: I tried ScheduledExecutorService in past to avoid thread sleep but it didn't work in this case. Basically, I was not getting any stats. 

**Why we are having thread sleep**
If you will check the design they are calling thread sleep on a runnable method body the runnable task is then executedin a separate thread which is responsible for reporting metrics, hence it will not block the other routines. The design is async since we are using separate threads for processing the dump.
In the past Zimbra avoid using their TaskSchedular class for the very same reason.


**To summarise the design**: We have Two classes StatsDumper and StatsDumperPrometheus (which inherits a few functionalities from StatsDumper but has completely different use-case)

**Why we are using callable and not runnable** -> We have a few exceptions that we want to handle, from the documentation I confirm that Runnable cannot throw checked exceptions, Runnable is a functional interface and its run method does not accept parameters. We want to pass varying parameters to collect stats from different activity trackers in our case. Callable is an improved version of runnable(from Baeldung).

For me, this design is okay since it already fulfills the requirement that is on our table.

I can extract a class to make it more DRY as suggested but we will end up doing the same thing we are doing right now. I can explain why if you want in a quick call as always :P

File: common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java [hunk 3]
```diff
@@ -0,0 +1,181 @@
...
+            } catch (InterruptedException e) {
+              ZimbraLog.perf.info("Stats thread interrupted: %s", dataSource.getFilename(), e);
+            }
+            if (Thread.currentThread().isInterrupted()) {
+              ZimbraLog.perf.info("Stats thread was interrupted: %s", dataSource.getFilename());
+            }
+          }
+        };
+    new Thread(STATS_GROUP, r, dataSource.getFilename()).start();
+  }
```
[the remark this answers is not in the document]
keshavbhatt: I have extracted a Runnable, we are using it to create instances and passing them to thread.

common/src/java-test/com/zimbra/common/stats/PrometheusFormatterTest.java
```diff
@@ -0,0 +1,85 @@
...
+
+    String expected =
+        "soap_exec_count {command=\"NoOpRequest\"} 2\n"
+            + "soap_exec_ms_avg {command=\"NoOpRequest\"} 0\n";
+
+    assertEquals(expected, buf.toString());
+  }
+
+  @Test
+  public void ShouldCleanHeaderWhenCalledSanitizeHeader() {
```
frisonisland, line 27: The convention is to use lowerCamelCase for methods

common/src/java-test/com/zimbra/common/stats/PrometheusFormatterTest.java
```diff
@@ -0,0 +1,85 @@
...
+            + "threads_SocketAcceptor 0\n"
+            + "threads_Thread 2\n"
+            + "threads_Timer 5\n"
+            + "threads_btpool 0\n"
+            + "threads_pool 0\n";
+    assertEquals(expected, buf.toString());
+  }
+
+  @Test
+  public void AddTimestampWhenCalledAddTimeStamp() {
```
frisonisland, line 78: Also tests should start with "should" convention (we decided together on this I believe).

common/src/java-test/com/zimbra/common/stats/PrometheusFormatterTest.java
keshavbhatt: my bad :smile_cat:

common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java
```diff
@@ -0,0 +1,77 @@
...
+
+  /**
+   * Writes log buffer to metrics file
+   *
+   * @param logBuffer the logBuffer containing processed data
+   * @throws IOException exception that could occur while writing data to file
+   */
+  private void writeLogBuffer(final StringBuilder logBuffer) throws IOException {
+    File file = getFile();
+    try (FileWriter writer = new FileWriter(file, false)) {
```
frisonisland, line 73: Any reason why we are writing not appending to file?

common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java
frisonisland: @keshavbhatt so are you saying that using an ExecutorService with Callable<V> didn't work?

frisonisland (changes requested):
Just a couple of comments (Java Doc), and a couple of questions left about ExecutorService and file writing

common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java
keshavbhatt: Yes, this is because we don't want to accumulate values in these files unlike we were doing in csv. Node exporter read data from here and look for new data everytime, dont care about old one.

common/src/java/com/zimbra/common/stats/StatsDumperPrometheus.java
keshavbhatt: @Polpetta Thanks for the deep investigation. We will do the refactoring in the next interation. We can have a quick call to specify what to use and what not.

frisonisland (approved):

Polpetta (approved):

PR #39, opened 2022-05-23
Feat/consul smtp

frisonisland:
### Goal

Define Consul MTA Upstream dependency and use it to send emails.

### What's changed

* mailbox upstream dependencies
* SmtpConfig.java and SmtpTransport.java classes to use static upstream host and port.

Linting has changed a lot of code, **to help you in the review this is what I changed**:

* carbonio-mailbox.hcl file
* SmtpTransport connect() overridden method + protocolConnect method conditional host and port (lines 158-164)
* SmtpConfig lines 15-17


store/src/java/com/zimbra/cs/mailclient/smtp/SmtpTransport.java
```diff
@@ -96,216 +95,239 @@
...
-            port =  PropUtil.getIntSessionProperty(session, "mail." + protocol + ".port",
-                    ssl ? SmtpConfig.DEFAULT_SSL_PORT : SmtpConfig.DEFAULT_PORT);
-        }
+    if (auth && (user == null || passwd == null)) {
+      return false;
+    }
 
-        // mail.protocol.host is examined in Service class
+    if (port < 0) {
+      port = ssl ? SmtpConfig.DEFAULT_SSL_PORT : SmtpConfig.DEFAULT_PORT;
```
keshavbhatt, line 159: is this check necessary here? 
I saw both `SmtpConfig.DEFAULT_SSL_PORT` and `SmtpConfig.DEFAULT_PORT;` are assigned explicitly to have the same value in the SmtpConfig.java class.
If we don't care about SSL for internal transport using consul, we can consider cleaning this.

keshavbhatt (approved):

store/src/java/com/zimbra/cs/mailclient/smtp/SmtpTransport.java
frisonisland: I think it's logically correct to do so. SmtpTransport doesn't know that SSL and non-SSL port are the same.
Right now SmtpConfig knows that, because we introduced Consul. I'm more concerned about protocol variable to be honest, because the underlying service is plain smtp

store/src/java/com/zimbra/cs/mailclient/smtp/SmtpTransport.java
```diff
@@ -96,216 +95,239 @@
...
+   * Constructs a new {@link SmtpTransport}.
+   *
+   * @param session JavaMail session
+   * @param urlname URL name
+   * @param ssl true to enable SSL, otherwise false
+   */
+  SmtpTransport(Session session, URLName urlname, boolean ssl) {
+    super(session, urlname);
+    this.ssl = ssl;
+    protocol = ssl ? "smtps" : "smtp";
```
frisonisland, line 130: Doubtful about this, because the underlying service accepts smtp and not smtps AFAIK.

frisonisland: This needs further investigation, because there is also SmtpsTransport class that extends this one. That's why there are both options for smtp and smtps. But it seems smtps is not used anywhere explicitly

PR #42, opened 2022-05-26
Feat/consul smtp

frisonisland:
### Goal

Add Consul MTA upstream dependency to mailbox + default smtp config when attributes are empty

### What's changed

* **carbonio-mailbox.hcl**
* **SmtpConfig.java** with new defaults
* **SmtpTransport.java** uses new default host
* added **TODO to other files** to track what we may need to change in the future


M0Rf30 (approved):

store/src/java/com/zimbra/cs/account/ZAttrServer.java
```diff
@@ -48618,25 +48620,25 @@ public Map<String,Object> unsetSmtpHostname(Map<String,Object> attrs) {
...
      *
      * @see #getSmtpPortAsString()
      *
-     * @return zimbraSmtpPort, or 25 if unset
+     * @return zimbraSmtpPort, or default if unset
      */
     @ZAttr(id=98)
     public int getSmtpPort() {
-        return getIntAttr(Provisioning.A_zimbraSmtpPort, 25, true);
+        return Integer.parseInt(getSmtpPortAsString());
```
frisonisland, line 48631: Note: this is not handled on purpose. If value is different from an integer, then underlying config is wrong and an exception will be raised.

store/src/java/com/zimbra/cs/util/JMSession.java
```diff
@@ -316,6 +316,8 @@ private static Properties getJavaMailSessionProperties(Server server, Domain dom
             throw new MessagingException(msg);
         }
         Properties props = new Properties(sSession.getProperties());
+        //TODO: when prop is null set property throws null pointer exception,
```
keshavbhatt, line 319: This task is for the future?

keshavbhatt (approved):
Suggested a few changes, everything else looks good.

store/src/java/com/zimbra/cs/account/ZAttrDomain.java
```diff
@@ -22094,25 +22095,25 @@ public Map<String,Object> unsetSmtpHostname(Map<String,Object> attrs) {
     /**
      * the SMTP server port to connect to when sending mail
      *
-     * <p>Use getSmtpPortAsString to access value as a string.
+     * <p>Internally uses getSmtpPortAsString to access value as a string.
      *
      * @see #getSmtpPortAsString()
      *
-     * @return zimbraSmtpPort, or -1 if unset
+     * @return zimbraSmtpPort, or default if unset
```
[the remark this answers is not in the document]
frisonisland: You're right, will update! Thanks

store/src/java/com/zimbra/cs/util/JMSession.java
frisonisland: It is, but it is more of a "watch out", I'm leaving things unchecked on purpose here.

Polpetta (approved):

PR #43, opened 2022-05-31
[CO-191] Cookie Safety

keshavbhatt:
### What's new:
**SameSiteAttribute** Utility class to set SameSite attribute of the Set-Cookie HTTP response header.

### What has changed:
Both `ZM_AUTH_TOKEN` and `ZM_ADMIN_AUTH_TOKEN` cookies have the `Secure` attribute set to `true` by default and the `SameSite` Attribute is set to `Lax`(https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite#lax) as requested.

### DIFF
--- a/common/src/java-test/com/zimbra/common/util/SameSiteAttributeTest.java
+++ b/common/src/java-test/com/zimbra/common/util/SameSiteAttributeTest.java
@@ -1,0 +1,17 @@
+package com.zimbra.common.util;
+
+import static org.junit.Assert.assertThat;
+import static org.junit.matchers.JUnitMatchers.containsString;
+
+import org.junit.Test;
+
+public class SameSiteAttributeTest {
+
+  @Test
+  public void addSameSiteAttributeToSetCookieHeaderWhenCalledAddSameSiteAttribute() {
+    MockHttpServletResponse response = new MockHttpServletResponse();
+    response.setHeader("Set-Cookie", "id=a3fWa; Max-Age=2592000");
+    SameSiteAttribute.addSameSiteAttribute(response, "Lax");
+    assertThat(response.getHeader("Set-Cookie"), containsString("SameSite=Lax"));
+  }
+}
--- a/common/src/java-test/com/zimbra/common/util/SameSiteAttributeTest.java
+++ b/common/src/java-test/com/zimbra/common/util/SameSiteAttributeTest.java
@@ -1,7 +1,6 @@
 package com.zimbra.common.util;
 
-import static org.junit.Assert.assertThat;
-import static org.junit.matchers.JUnitMatchers.containsString;
+import static junit.framework.Assert.assertEquals;
 
 import org.junit.Test;
 
@@ -12,6 +11,6 @@ public class SameSiteAttributeTest {
     MockHttpServletResponse response = new MockHttpServletResponse();
     response.setHeader("Set-Cookie", "id=a3fWa; Max-Age=2592000");
     SameSiteAttribute.addSameSiteAttribute(response, "Lax");
-    assertThat(response.getHeader("Set-Cookie"), containsString("SameSite=Lax"));
+    assertEquals("id=a3fWa; Max-Age=2592000; SameSite=Lax", response.getHeader("Set-Cookie"));
   }
 }
--- a/common/src/java/com/zimbra/common/util/SameSiteAttribute.java
+++ b/common/src/java/com/zimbra/common/util/SameSiteAttribute.java
@@ -10,24 +10,28 @@ import javax.servlet.http.HttpServletResponse;
  *     attribute allows us to declare if the cookie should be restricted to a first-party or
  *     same-site context
  */
-public class SameSiteAttribute {
+class SameSiteAttribute {
+
+  private static final String SET_COOKIE_HEADER = "Set-Cookie";
+
+  private SameSiteAttribute() {}
 
   /**
    * @param response the response object containing the Set-Cookies header
    * @param sameSiteValue value to be set for SameSite attribute
    */
   public static void addSameSiteAttribute(HttpServletResponse response, String sameSiteValue) {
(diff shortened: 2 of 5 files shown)


frisonisland (approved):

Polpetta (approved):

common/src/java-test/com/zimbra/common/util/SameSiteAttributeTest.java
```diff
@@ -0,0 +1,17 @@
...
+import org.junit.Test;
+
+public class SameSiteAttributeTest {
+
+  @Test
+  public void addSameSiteAttributeToSetCookieHeaderWhenCalledAddSameSiteAttribute() {
+    MockHttpServletResponse response = new MockHttpServletResponse();
+    response.setHeader("Set-Cookie", "id=a3fWa; Max-Age=2592000");
+    SameSiteAttribute.addSameSiteAttribute(response, "Lax");
+    assertThat(response.getHeader("Set-Cookie"), containsString("SameSite=Lax"));
```
frisonisland, line 15: Wouldn't it be better to check value is "equal to" instead of "contains"? 
I'm saying this because what if the ";" symbol is missing before SameSite=Lax ?

frisonisland (changes requested):

common/src/java-test/com/zimbra/common/util/SameSiteAttributeTest.java
keshavbhatt: we can not check for equality since the Set-Cookie attribute value constains more than that. 


> I'm saying this because what if the ";" symbol is missing before SameSite=Lax ?

That won't happen, since we are adding `;` while re-constructing the new header in the addSameSiteAttribute method of the SameSiteAttribute utility class.

keshavbhatt: > but we control the input

The input(key value pair are managed/added by the standard method of HttpServletResponse Class) and they take cares of the correct formatting while appending headers to the response object. They won't add extras `;` unless there are more attributes to appeded to the header. I checked that and I confirm. So don't worry :)

frisonisland: Sorry, maybe I did not explain myself correctly. I mean that the header, after addSameSiteAttribute, should be equal to
"id=a3fWa; Max-Age=2592000, SameSite=Lax". Checking this we also confirm it's set as per RFC. Isn't it?
I'm not worried about HttpResponse adding extra ";"! I'm just talking about this class. 
See what I did

image

keshavbhatt: Made the test case more specific to check the formatting. 
Thanks for drawing attention to this.

keshavbhatt: > so I expect to have something like id=a3fWa; Max-Age=2592000; SameSite=Lax;

proper format is : `id=a3fWa; Max-Age=2592000; SameSite=Lax`
no `;` at end unless we have more k/v pairs to add.

common/src/java/com/zimbra/common/util/SameSiteAttribute.java
```diff
@@ -0,0 +1,37 @@
...
+    for (String header : headers) {
+      if (firstHeader) {
+        response.setHeader(
+            SET_COOKIE_HEADER, String.format("%s; %s", header, "SameSite=" + sameSiteValue));
+        firstHeader = false;
+        continue;
+      }
+      response.addHeader(
+          SET_COOKIE_HEADER, String.format("%s; %s", header, "SameSite=" + sameSiteValue));
+    }
```
frisonisland, line 35: I checked HttpServletResponse interface and it seems that the **getHeaders()** method returns a list of header values (not headers). Also the addHeader() method will add values to a header, so this is what I would do:
```suggestion
    if (!(Objects.isNull(response.getHeader(SET_COOKIE_HEADER)))) {
      response.addHeader(SET_COOKIE_HEADER, "SameSite=" + sameSiteValue);
    }
```
The result of my suggestion should be the value "SameSite=something" added to Set-Cookie header.
**I fear the solution you proposed may add a "SameSite=Lax" to each value of the header** (e.g.: Set-Cookie: id=1; SameSite=Lax; Max-Value=b; SameSite=Lax).
**Unfortunately the Unit Test is not able to show and test if this is the case, because MockHttpServletResponse does not implement all methods the right way.** If possible I would check with a "real" HttpServletResponse (IT)

frisonisland (changes requested):

common/src/java/com/zimbra/common/util/SameSiteAttribute.java
keshavbhatt: I will test it to see if it works.

keshavbhatt: It doesn't work. 
> I fear the solution you proposed may add a "SameSite=Lax" to each value of the header 
It won't :) we have the checks are in place. 
The provided solution is tested and fulfills the requirements already.

frisonisland: > It won't :) we have the checks are in place. The provided solution is tested and fulfills the requirements already.
There is only a test with a MockHttpServletResponse.
To fulfill the requirements we need a test with a real HttpResponse and a Set-Cookie header with multiple values

frisonisland: We checked together that the implementation result is as expected!
It will just be a plus if we could test SameSiteAttribute class against jetty Response class! Thanks :)

frisonisland (approved):
Everything works as expected. As I said it would just be a plus to test against a real server (spin-up) or just against Jetty Response class

PR #46, opened 2022-06-07
Mailboxd access consul kv

flanciotti:
changes to allow mailboxd, executed by user zextras, to access Key-Value store in Consul


Polpetta (approved):

M0Rf30 (approved):

frisonisland (approved):
Ok to me. I would only be careful in the future when we add more policies rules.
It's still a special case that zextras is in carbonio-mailbox group.
This is not true for all other services!

keshavbhatt (approved):

PR #47, opened 2022-06-07
Fix/carbonio auth fallback

frisonisland:
## Goal
Fallback should not happen when using carbonio advanced auth mechanism.

### What's changed
**AuthMechanism:** 
- rewrite custom:carbonioAdvanced -> carbonioAdvanced
- add isDefaultAuth method (checks whether is carbonioAdvanced or zimbra auth)

**LdapProvisioning:**
- line 6388 (sry): instead of checking on isZimbraAuth() it will check on isDefaultAuth() method.
The idea is: if auth is default then do not fallback

### DIFF
--- a/store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
+++ b/store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
@@ -114,9 +114,13 @@ public abstract class AuthMechanism {
 
     final String carbonioAdvancedMech =
         AuthMech.custom.name() + ":" + AuthMech.carbonioAdvanced.name();
+
+    // if custom:carbonioAdvanced skip to switch case
+    if (Objects.equals(authMechStr, carbonioAdvancedMech)) {
+      authMechStr = AuthMech.carbonioAdvanced.name();
+    }
     if (authMechStr.startsWith(AuthMech.custom.name() + ":")) {
-      return new CustomAuth(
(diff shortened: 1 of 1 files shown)


store/src/java/com/zimbra/cs/account/auth/AuthMechanism.java
```diff
@@ -28,438 +21,473 @@
...
-        ZimbraAuth zimbraAuth = new ZimbraAuth(AuthMech.zimbra);
-        zimbraAuth.doAuth(prov, domain, acct, password, authCtxt);
-    }
+    final String carbonioAdvancedMech =
+        AuthMech.custom.name() + ":" + AuthMech.carbonioAdvanced.name();
 
-    public boolean isZimbraAuth() {
-        return false;
+    // if custom:carbonioAdvanced skip to switch case
+    if (Objects.equals(authMechStr, carbonioAdvancedMech)) {
```
frisonisland, line 119: Fellow reviewers, this has been changed.
To trigger carbonioAdvanced auth we have three ways:

1. Install carbonioAdvanced and leave zimbraAuthMech empty -> switch case
2. zimbraAuthMech equal to carbonioAdvanced -> switch case
3. zimbraAuthMech equal to custom:carbonioAdvanced -> **this if condition will rewrite custom:carbonioAdvanced to carbonioAdvanced and will end up in switch case**, following the right logic.

**Without the rewrite we will end up with a standard CustomAuth, but it will always fallback on ldap and it's not the wanted behaviour.** Thanks

Polpetta (approved):

PR #48, opened 2022-06-08
chore: [CO-288] add support for ngnix templates override

keshavbhatt:
### What's new
The custom template override location that proxyConfGen will  use is set to `/opt/zextras/conf/nginx/templates_custom`

**Config generation flow:** If one or more template overrides are found in the custom templates location(`/opt/zextras/conf/nginx/templates_custom`), the proxyconfgen util will now use those templates to generate the config and will override the default ones that exist in the default template location (`/opt/zextras/conf/nginx/templates`) at the proxyxonfgen step while services start or by manual invocation of the tool.

**Warning header:** A warning header is written to all generated configuration files.

NOTE: The above flow won't work if a custom template path is specified explicitly to the proxyconfgen command using its -t flag. This ensures the backward compatibility of proxyconfgen’s behavior. 

### Changes in other repositories:

- **carbonio-nginx-conf** create template override dir  ->  https://github.com/Zextras/carbonio-nginx-conf/pull/22

- **carbonio-core-utils** zmfixperm support  ->  https://github.com/Zextras/carbonio-core-utils/pull/16

### DIFF
--- a/store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
+++ b/store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
@@ -27,6 +27,8 @@ import java.net.Inet4Address;
 import java.net.Inet6Address;
 import java.net.InetAddress;
 import java.net.UnknownHostException;
+import java.nio.file.Files;
+import java.nio.file.Paths;
 import java.util.ArrayList;
 import java.util.Collections;
 import java.util.HashMap;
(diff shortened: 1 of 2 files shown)


frisonisland (changes requested):
I have to ask to add tests as proof of the wanted behavior.
Example: 
- test output path
- test pattern match

keshavbhatt:
> * test output path

I don't get what you mean? output path is a constant what do we want to test about it? Sorry if I am not able to decode you again :crying_cat_face: 

> * test pattern match

can you be more specific? I did not add any new patterns in these changes. :smiling_face_with_tear:

Polpetta:
I think @frisonisland wants the tests that:
* given the template and compiled it you have an expected result
* you want to test the regex and possible edge cases
* you want to check that the warning is printed when necessary and that fulfills behavior

keshavbhatt:
> I think @frisonisland wants the tests that:
> 
> * given the template and compiled it you have an expected result

I will be more than happy to do that, but I have not touched the formatting of generated files they are the same and installed at the same location.

> * you want to test the regex and possible edge cases

No changes were made to affect the formatting. 

> * you want to check that the warning is printed when necessary and that fulfills behavior
 
warning is printed only when usingCustomTemplateOverride is set to true.

adding any of the tests mentioned above requires a call to expandTemplate method. So basically we are talking about writing the whole test suit for the proxyxconfgen utility. We will do that in another ticket. since we need these changes in 4.0.10

flanciotti (approved):

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -395,9 +400,16 @@ public static String getDefaultClientCertCaPath() {
...
     String configFilePath = configFile.getAbsolutePath();
 
+    final String templateFileName = templateFile.getName();
+    final String overrideTemplateFilePath =
+        OVERRIDE_TEMPLATE_DIR + File.separator + templateFileName;
+    final boolean usingCustomTemplateOverride =
+        !hasCustomTemplateLocationArg && Files.exists(Paths.get(overrideTemplateFilePath));
+    final String templateFilePath =
+        usingCustomTemplateOverride ? overrideTemplateFilePath : templateFile.getAbsolutePath();
+
```
frisonisland, line 412: Hey Keshav, to answer your questions about testing: in the expandTemplate method **we added a check on custom template argument to infer the templateFilePath.** I think we can at least add a test about this behavior.
Everything else should have already been tested (file content, etc.) long time ago :wink: 

**I would suggest to create another method (e.g.: getTemplateFilePath) that returns the template directory output and test its value based on hasCustomTemplateLocationArg condition**.
Hope it's clear! :slightly_smiling_face:

frisonisland (changes requested):

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
keshavbhatt: Okay, my lord will do that.

frisonisland (approved):

PR #51, opened 2022-06-10
[CO-300] proxy 5xx errors to 404 in previewServlet

keshavbhatt:
key change:

- https://github.com/Zextras/carbonio-mailbox/pull/51/files#diff-af5013b1979149a28c4275ed38534e3d75c5a947a7253b54973229ab94ea774dR356

- https://github.com/Zextras/carbonio-mailbox/pull/51/files#diff-d616bfd8f0383718e7c8a758c940f71d19360e468411efe19b21dc6229b8f055R122

### DIFF
--- a/store/src/java-test/com/zimbra/cs/service/PreviewServletTest.java
+++ b/store/src/java-test/com/zimbra/cs/service/PreviewServletTest.java
@@ -118,6 +118,14 @@ public class PreviewServletTest {
     assertEquals(401, mockResponse.getStatus());
   }
 
+  @Test // CO-300
+  public void shouldProxy5xxErrorsWith404() throws IOException {
+    final PreviewServlet previewServlet = new PreviewServlet();
+    MockHttpServletResponse mockResponse = new MockHttpServletResponse();
+    previewServlet.sendError(mockResponse, 500, "Internal ServerError");
+    assertEquals(HttpServletResponse.SC_NOT_FOUND, mockResponse.getStatus());
+  }
+
   @Test
   public void shouldProvideCorrectPreviewQueryParametersWhenCalledParseQueryParameters()
       throws Exception {
--- a/store/src/java-test/com/zimbra/cs/service/PreviewServletTest.java
+++ b/store/src/java-test/com/zimbra/cs/service/PreviewServletTest.java
@@ -1,7 +1,9 @@
 package com.zimbra.cs.service;
 
 import static org.junit.Assert.assertEquals;
+import static org.mockito.Mockito.atLeastOnce;
 import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.verify;
 import static org.mockito.Mockito.when;
 
 import com.google.common.collect.Maps;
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/service/PreviewServlet.java
```diff
@@ -338,12 +339,29 @@ void respondWithSuccess(HttpServletResponse resp, BlobResponseStore blobResponse
...
 
+  /**
+   * This override method is to proxy 5xx error with 404
+   *
+   * @param resp the {@link HttpServletResponse} object
+   * @param errCode error code for {@link HttpServletResponse}
+   * @param reason message string for {@link HttpServletResponse}
+   * @throws IOException while sendError
+   */
+  void sendError(HttpServletResponse resp, int errCode, String reason) throws IOException {
```
frisonisland, line 356: **I would consider running the test against the doGet() method instead of sendError()**, just because that is the API entrypoint.
Also this way  there is no need to make sendError() package private for the sake of testing.

store/src/java-test/com/zimbra/cs/service/PreviewServletTest.java
```diff
@@ -118,6 +118,14 @@ public void shouldReturn401WhenCheckAuthTokenFromCookieOrRespondWithErrorIfCooki
     assertEquals(401, mockResponse.getStatus());
   }
 
+  @Test // CO-300
+  public void shouldProxy5xxErrorsWith404() throws IOException {
```
frisonisland, line 122: I would consider running the test against the doGet() method instead of sendError(), just because that is the API entrypoint.
Also this way there is no need to make sendError() package private for the sake of testing.

store/src/java/com/zimbra/cs/service/PreviewServlet.java
```diff
@@ -338,12 +339,29 @@ void respondWithSuccess(HttpServletResponse resp, BlobResponseStore blobResponse
...
+    if (resp.getStatus() / 100 == 5 || errCode / 100 == 5) {
```
frisonisland, line 357: Consider using HttpStatus.isServerError() (org.eclipse.jetty.http) for clean code

frisonisland (approved):
Left just some suggestions! :+1:

store/src/java-test/com/zimbra/cs/service/PreviewServletTest.java
keshavbhatt: The problem with that is you won't be able to simulate a real internal server error. 
I have a wrapper method that I use to send any error message produced by the API. 
So, I think it's okay to test that wrapper method, cause all the errors to have to pass through it.

frisonisland: I see, but a scenario could be: we implemented the sendError() method, but did not used it correctly /always where required.
We could implement a test for the doGet(), by mocking PreviewClient with mockito and make it return 500, and see how the response is handled.

Problem is that PreviewClient is not assigned in the Servlet constructor, so there is no way to mock it right now.
We should find/study a way to inject instances on the Servlet!

keshavbhatt: I have added one test. That should cover the scenario.

PR #53, opened 2022-06-16
chore: [Co-303] enable tests with coverage in CI

frisonisland:
### Goal

Add Junit report and JaCoCo report coverage

### What's changed

- added **jacoco-common.xml** to load and download JaCoCo plugin
- added **junit-common.xml** to load and download Junit plugin
- **removed old Emma plugin** coverage build-common.xml
- **wrapped junit tests with jacoco** agent in build-common.xml
- add ant target to merge multiproject JaCoCo coverage
- Jenkinsfile: jacoco and junit publish reports 
- Jenkinsfile: run all tests (unit & IT) when TEST param enabled

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -54,6 +54,7 @@ pipeline {
                    test-all-coverage-plough-through
                """
             publishCoverage adapters: [jacocoAdapter('build/coverage/merged.xml')]
+            junit '**/build/test/output/*.xml'
           }
         }
         stage("Publish to maven") {
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -46,6 +46,16 @@ pipeline {
                 stash includes: 'staging/**', name: 'staging'
             }
         }
+        stage("Test with coverage \(allow failure\)") {
+          steps {
+            sh """
+                   ANT_RESPECT_JAVA_HOME=true JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/ ant \
+                   -propertyfile build.properties \
+                   test-all-coverage-plough-through
+               """
+            publishCoverage adapters: [jacocoAdapter('build/coverage/merged.xml')]
+          }
+        }
         stage("Publish to maven") {
             when {
                 buildingTag()
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -49,7 +49,7 @@ pipeline {
         stage("Test with coverage (allow failure)") {
           steps {
             sh """
-                   ANT_RESPECT_JAVA_HOME=true JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/ ant \
+                   ANT_RESPECT_JAVA_HOME=true JAVA_HOME=/usr/lib/jvm/java-11-openjdk-amd64/ ant -d \
                    -propertyfile build.properties \
                    test-all-coverage-plough-through
                """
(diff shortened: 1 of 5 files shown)


keshavbhatt (approved):

M0Rf30 (approved):
Everything looks good to me, my main suggestion is related to formatting of build.xml and build-common.xml. In the past I used a specific xml formatter. If the tool is the same across all the team members and I'm the singularity I will adopt your tool, otherwise I can submit a commit, queued to this and only for the mentioned files. Let me know

frisonisland:
> Everything looks good to me, my main suggestion is related to formatting of build.xml and build-common.xml. In the past I used a specific xml formatter. If the tool is the same across all the team members and I'm the singularity I will adopt your tool, otherwise I can submit a commit, queued to this and only for the mentioned files. Let me know

Hey @M0Rf30 we are using this config: https://github.com/Zextras/carbonio-mailbox/blob/main/.pre-commit-config.yaml
We can add an xml linter if you want. 
Right now I'm using the IDE formatter (should be this: https://google.github.io/styleguide/intellij-java-google-style.xml)

M0Rf30:
> > Everything looks good to me, my main suggestion is related to formatting of build.xml and build-common.xml. In the past I used a specific xml formatter. If the tool is the same across all the team members and I'm the singularity I will adopt your tool, otherwise I can submit a commit, queued to this and only for the mentioned files. Let me know
> 
> Hey @M0Rf30 we are using this config: https://github.com/Zextras/carbonio-mailbox/blob/main/.pre-commit-config.yaml We can add an xml linter if you want. Right now I'm using the IDE formatter (should be this: https://google.github.io/styleguide/intellij-java-google-style.xml)

got it. I will sync with you and with the suggested style. thanks

Jenkinsfile
```diff
@@ -46,6 +47,20 @@ pipeline {
                 stash includes: 'staging/**', name: 'staging'
             }
         }
+        stage("Test all with coverage (allow failure)") {
```
Polpetta, line 50: If you want you can mark the build "stable" with % of test failed, so that from now on we try to diminish that number

Polpetta (changes requested):

Jenkinsfile
frisonisland: > If you want you can mark the build "stable" with % of test failed, so that from now on we try to diminish that number

I didn't do it because I don't have an idea of how much % we need, I think we should reason about that and do it in next iterations

frisonisland:
> Also, I'm throwing an idea here, how much difficult would be to use Junit5 here?

A lot more work, I was taking care about it here: https://github.com/Zextras/carbonio-mailbox/tree/fix/working-tests
Reason is that we have some issues in downloading transitive dependencies and I found we need to update dependency manager logic (Ivy) to fully use Junit5.
Of course feel free to work on that branch and improve it!

Jenkinsfile
frisonisland: We solved this IRL :slightly_smiling_face: . Junit plugin does not allow for stable/unstable build based on % failure. We discussed several options we can use in the future to achieve this goal. 
Please @Polpetta mark it as solved if you're satisfied :wink:

Polpetta:
> > Also, I'm throwing an idea here, how much difficult would be to use Junit5 here?
> 
> A lot more work, I was taking care about it here: https://github.com/Zextras/carbonio-mailbox/tree/fix/working-tests Reason is that we have some issues in downloading transitive dependencies and I found we need to update dependency manager logic (Ivy) to fully use Junit5. Of course feel free to work on that branch and improve it!

Cool cool!

Polpetta (approved):

PR #58, opened 2022-06-27
chore: [CO-312] utilise the generator to create classes from attrs.xml

keshavbhatt:
These changes are the outcome of utilizing the `generate-getter` target on the store module.

**Test:** (Tested on a fresh system, single server)
- Bootstrap is happy
- Services are running fine
- Provisioning works fine with at least a few random tests


Polpetta:
I don't have a clue of how to check it. What's the procedure in order to generate all the code in this pr?

keshavbhatt:
> I don't have a clue of how to check it. What's the procedure in order to generate all the code in this pr?

Sorry for leaving out the details: ant generate-getters in the store module.

The majority of the changes are there because we didn't use the utility to generate the code for long and had to manually add the code with our hands.

frisonisland:
> I don't have a clue of how to check it. What's the procedure in order to generate all the code in this pr?

Maybe we can add these info to the README, what do you think @keshavbhatt ?

keshavbhatt:
> > I don't have a clue of how to check it. What's the procedure in order to generate all the code in this pr?
> 
> Maybe we can add these info to the README, what do you think @keshavbhatt ?

Of course, we should add a section.

store/src/java/com/zimbra/cs/account/ZAttrServer.java
```diff
@@ -33917,13 +34036,13 @@ public Map<String,Object> unsetMtaSmtpdTlsLoglevel(Map<String,Object> attrs) {
      *
      * <p>Valid values: [export, low, medium, high, null]
      *
-     * @return zimbraMtaSmtpdTlsMandatoryCiphers, or ZAttrProvisioning.MtaSmtpdTlsMandatoryCiphers.medium if unset and/or has invalid value
+     * @return zimbraMtaSmtpdTlsMandatoryCiphers, or ZAttrProvisioning.MtaSmtpdTlsMandatoryCiphers.high if unset and/or has invalid value
```
keshavbhatt, line 34039: It seems like someone manually changed `MtaSmtpdTlsMandatoryCiphers` value to medium in zattrs source files and did nothing on the attrs.xml side. The generator used the attrs.xml default value hence the change.
Anyway, Postfix's official documentation(https://www.postfix.org/postconf.5.html#smtpd_tls_mandatory_ciphers) suggests using medium by default so we will set the default value in the attribute definition to medium

store/src/java/com/zimbra/cs/account/ZAttrServer.java
```diff
@@ -33637,13 +33756,13 @@ public Map<String,Object> unsetMtaSmtpdTlsCcertVerifydepth(Map<String,Object> at
      *
      * <p>Valid values: [export, low, medium, high, null]
      *
-     * @return zimbraMtaSmtpdTlsCiphers, or ZAttrProvisioning.MtaSmtpdTlsCiphers.export if unset and/or has invalid value
+     * @return zimbraMtaSmtpdTlsCiphers, or ZAttrProvisioning.MtaSmtpdTlsCiphers.high if unset and/or has invalid value
```
keshavbhatt, line 33759: this attribute has also picked the value defined in attrs.xml and dropped manually added export (aka medium for postfix releases past year 2015), we can change the default value in an attribute definition to medium here as well.

frisonisland (approved):
Changes are legit to me, we just need to take care and discuss about some changed values that were not reflected in attrs.xml file.

Edit: checked that values in attrs.xml are what we want and they are a result of security hardening :slightly_smiling_face:

M0Rf30 (approved):

store/src/java/com/zimbra/cs/account/ZAttrServer.java
keshavbhatt: It was discovered that the values specified in attrs.xml were correct as a result we will be maintaining them as they are.

Polpetta (approved):

store/src/java/com/zimbra/cs/account/ZAttrServer.java
keshavbhatt: The information will be included to the Readme in a separate PR.

store/src/java/com/zimbra/cs/account/ZAttrServer.java
keshavbhatt: It was discovered that the values specified in attrs.xml were correct as a result we will be maintaining them as they are.

PR #66, opened 2022-07-21
fix(proxy): [CO-321] server side TLS handshake

keshavbhatt:
What has changed:
- These changes will let the proxy use a new set of algorithms while doing TLS/SSL handshake.
- remove explicit TLSV1.3 protocol usage for mailboxd this prevents random handshake failures(SSL alert 80) caused due to issue with jetty on JDK11 (bug JDK-8213202)

### DIFF
--- a/store/conf/attrs/attrs.xml
+++ b/store/conf/attrs/attrs.xml
@@ -3421,7 +3421,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
 </attr>
 
 <attr id="640" name="zimbraReverseProxySSLCiphers" type="string" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" requiresRestart="nginxproxy" since="5.0.5">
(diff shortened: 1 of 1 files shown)


frisonisland (approved):

M0Rf30 (approved):
Just to check if I understand correctly.
Do you used intermediate mozilla profile is it right?
Should be the safer choice, and about dropping TLS 1.3.
How this will work with jetty.xml.production? No TLS 1.3 at all?

keshavbhatt:
> Just to check if I understand correctly. Do you used intermediate mozilla profile is it right? Should be the safer choice, and about dropping TLS 1.3. How this will work with jetty.xml.production? No TLS 1.3 at all?

We are not dropping TLS1.3(we are not excluding the protocol), Jetty is still able to use TLSV1.3 the value is set through the mailboxd_java_options in localconfig.

M0Rf30 (approved):

PR #68, opened 2022-07-29
Code cleanup: ZDesktop

keshavbhatt:
**Key changes:** 

* remove zdesktop_local_account_id entry from **locale resources**
* remove zdesktop_local_account_id entry from **local config class (LC)**
* remove **traces of zdesktop** from the codebase, get rid of warnings due to legacy code
* code cleanup that includes
    - use parent class ZAttrProvisioning to access static members instead of Provisioning
    - use standard charset definition from class StandardCharsets instead of self-declared version
    - change array designator (get rid of c style array declaration)
    - remove redundant access identifiers where required
    - remove unused fields
* add unit test for a few methods of ContentServlet and AccountUtil class 

**How to check the changes:** suggest using your IDE(ignoring imports and formatting) to check the diffs instead github's webview


store/src/java-test/com/zimbra/cs/util/AccountUtilTest.java
```diff
@@ -0,0 +1,89 @@
...
+
+    Account account = prov.get(Key.AccountBy.name, GLOBAL_TEST_ACCOUNT_NAME);
+    Server server = prov.getServer(account);
+    server.setServiceHostname("demo.zextras.com");
+    server.setMailPort(80);
+    assertEquals("http://demo.zextras.com:80/service/soap/", AccountUtil.getSoapUri(account));
+  }
+
+  @Test
+  public void shouldReturnWhenIsGalAccountCalled() throws ServiceException {
```
frisonisland, line 71: It seems we want to test that the method isGalSyncAccount just returns a value, by looking at the test name.
But since we are testing against True/False, I would suggest to split the test in two and make them specific:
- shouldReturnTrueWhenAccountHasGalAcctId
- shouldReturnFalseWhenAccountMissingGalAcctId

Also beware of test order execution: because with setGalAccountId we are modifying the state of a global accessible account, some conditions against isGalSyncAccount may return unwanted result if we add more tests in the future.

keshavbhatt: Renamed the test, and manually unset/set domain gal account Id so it won't be affected by global changes.

frisonisland (approved):

Polpetta (approved):

PR #70, opened 2022-08-01
chore: [CO-357] update mina-core to 2.1.6

frisonisland:
Updates apache mina-core to 2.1.6

### DIFF
--- a/store/src/java/org/apache/mina/transport/socket/nio/ZimbraSocketAcceptor.java
+++ b/store/src/java/org/apache/mina/transport/socket/nio/ZimbraSocketAcceptor.java
@@ -76,7 +76,7 @@ public final class ZimbraSocketAcceptor
 
   @Override
   protected void init(SelectorProvider selectorProvider) throws Exception {
-    selector = Selector.open();
+    this.init();
   }
 
   @Override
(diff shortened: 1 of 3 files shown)


store/src/java/org/apache/mina/transport/socket/nio/ZimbraSocketAcceptor.java
```diff
@@ -12,219 +12,224 @@
...
+
+  @Override
+  protected void init() throws Exception {
+    selector = Selector.open();
+  }
+
+  @Override
+  protected void init(SelectorProvider selectorProvider) throws Exception {
+    selector = Selector.open();
+  }
```
Polpetta, line 80: Sorry I don't get it why did you add this method if you don't use `selectorProvider`?

Polpetta (approved):
I've a question before approving

store/src/java/org/apache/mina/transport/socket/nio/ZimbraSocketAcceptor.java
frisonisland: @Polpetta I went for calling this.init()

keshavbhatt (approved):

PR #72, opened 2022-08-02
chore: [CO-333] deprecate FeatureTouchClientEnabled

keshavbhatt:
**Key changes:**
- mark FeatureTouchClientEnabled attribute as deprecated
- running the code generator also updated a few defaults we set a few days back for SSL ticket which was also intended but missed for some reasons


frisonisland (approved):
I think we have also TLS changes because we didn't run the generator in: https://github.com/Zextras/carbonio-mailbox/pull/66

Polpetta (approved):

PR #73, opened 2022-08-03
Fix/co 284 global admin fallback

frisonisland:
Fixes admin fallback logic.
Previously admin could always login with local credentials, now the decision is based on zimbraAuthFallbackToLocal value, like a regular user.


keshavbhatt (approved):

Polpetta (commented):
Can we test this change somehow?

frisonisland:
> Can we test this change somehow?

I have added a few test cases, but left them ignored. The problem is we need to test against the LdapProvisioning, but only MockProvisioning is available at the moment.

Polpetta (approved):

PR #77, opened 2022-08-10
feat: [CO-359] add ability to preview attachments in shared mailbox folders

keshavbhatt:
**Key changes:** 
- Preview servlet now can preview attachments that are part of shared email folders

### DIFF
--- a/store/src/java/com/zimbra/cs/service/PreviewServlet.java
+++ b/store/src/java/com/zimbra/cs/service/PreviewServlet.java
@@ -91,7 +91,18 @@ import org.eclipse.jetty.http.HttpStatus;
  */
 public class PreviewServlet extends ZimbraServlet {
 
-  private static final String SERVLET_PATH = "/preview";
+  public static final String SERVLET_PATH = "/preview";
+  public static final String THUMBNAIL_PDF_REGEX =
+      SERVLET_PATH + "/pdf/([a-zA-Z\\-:0-9]+|[0-9]+)/([0-9]+)/([0-9]*x[0-9]*)/thumbnail/?\\??(.*)";
+  public static final String PREVIEW_PDF_REGEX =
+      SERVLET_PATH + "/pdf/([a-zA-Z\\-:0-9]+|[0-9]+)/([0-9]+)/?((?=(?!thumbnail))(?=([^/\\n ]*)))";
+  public static final String THUMBNAIL_IMAGE_REGEX =
+      SERVLET_PATH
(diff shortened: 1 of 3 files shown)


store/src/java/com/zimbra/cs/service/PreviewServlet.java
```diff
@@ -124,19 +143,143 @@ static String getDispositionType(String requestUrl) {
...
+      HttpServletResponse resp, AuthToken authToken, String messageId, String part) {
+    return Try.of(
+        () -> {
+          HttpClientBuilder clientBuilder =
+              ZimbraHttpConnectionManager.getInternalHttpConnMgr().newHttpClient();
+          HttpProxyUtil.configureProxy(clientBuilder);
+          HttpGet getRequest =
+              new HttpGet(getContentServletResourceUrl(authToken, messageId, part));
+          HttpClient client =
+              encodeClientBuilderRequest(resp, authToken, clientBuilder, getRequest).get();
```
frisonisland, line 237: I have just a note **about passing down the response element as a parameter**: I would avoid it and **instead keep the response error logic at top level (in the doGet method)**.
I'm ok with the "sendError" method to manipulate the response, although I prefer the original resp.sendError of HttpServletResponse interface, but **passing down the response to each method just and modifying its status for each failure will inevitably result in confusion at some point** because we will not be able to trace where the error happened without proper debugging.

I would suggest instead in the doGet, for each method response, calling isFailure() and then set the response error.

Edit: the flow is:
doGet() -> getAttachment() -> encodeClientBuilderRequest() -> getMailHostUrl(), and to each of them the response is being passed.
But in the end **the response is not being used to make any http calls, but just to catch the failure and set the error.**
**Instead let the error flow up, until we catch it in the doGet()**

frisonisland (changes requested):
Just left a note about managing response error. 
My suggestion is to let the error flow up until the doGet() and only then set the error if returned status is failure.

store/src/java/com/zimbra/cs/service/PreviewServlet.java
frisonisland: Ok, solved IRL, we will take care in next related PR, the functionality is there

frisonisland (approved):

store/src/java/com/zimbra/cs/service/PreviewServlet.java
keshavbhatt: I agree with your proposal and I decided to add it to the next PR, which is also connected to something similar. I stated why in chat :wink:

Polpetta (approved):
LGTM. Remember next time to put finals into variables declaration if their pointers don't need to change :+1:

PR #79, opened 2022-08-12
fix: [CO-365] start preview servlet at mailbox boot

keshavbhatt:
**Goal:** The servlet is currently set to start when the first request is received. As a result of the startup time and preview processing time, the response to the initial request is delayed by a few more seconds.

**Key changes:** During mailbox startup, PreviewServlet is scheduled to start in the 13th position in the queue.


frisonisland (approved):
Makes sense to me

aheeva-yuliya:
LGTM!

aheeva-yuliya (approved):

PR #81, opened 2022-08-25
chore: new pipeline parameter to skip tests

keshavbhatt:
**Key changes:** 
- skip test with coverage step if parameter(SKIP_TEST_WITH_COVERAGE) is set checked or set to true
- file formatting

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,16 +1,17 @@
 pipeline {
     agent {
         node {
-            label 'carbonio-agent-v1'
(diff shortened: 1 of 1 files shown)


frisonisland (approved):

Polpetta (commented):
Are you sure you want to skip tests by default? It is better to keep the tests enabled by default and optionally disable it, imho

Polpetta (approved):
lol ok, I mis-read the variable name, sorry my bad! It's fine as it is

PR #82, opened 2022-08-26
Feature/co 362 proxy ip trusted

frisonisland:
### Goal

When an http call is received through the proxy, the original client Ip is stored in X-Forwarded-For header (see: https://www.nginx.com/resources/wiki/start/topics/examples/forwarded/).
**In order to propagate the original client ip to the context, the proxy ip address mus be trusted** by adding it to zimbraTrustedIp attribute (see: https://wiki.zimbra.com/wiki/Zimbra_Proxy_Manual:Miscellaneous_Topics).
**We want to automatically trust the proxy ip.**

### What's changed
I have modified the Server class to return the server hostname and its ip address by lazy-loading it, since a DNS resolution is required.
The static method ZimbraServlet#getTrustedIps now returns also the proxy ip address.

### DIFF
--- a/store/src/java-test/com/zimbra/cs/account/MockProvisioning.java
+++ b/store/src/java-test/com/zimbra/cs/account/MockProvisioning.java
@@ -37,8 +37,10 @@ import java.util.Collections;
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.Objects;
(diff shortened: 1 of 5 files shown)


Polpetta (approved):
LGTM

keshavbhatt (approved):
LGTM

PR #87, opened 2022-09-05
feat: zimbraAuthFallbackToLocal is by default set to TRUE

Polpetta:
* Regenerate attrs.xml based codebase

[CO-338]

### DIFF
--- a/store/conf/attrs/attrs.xml
+++ b/store/conf/attrs/attrs.xml
@@ -1470,7 +1470,7 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
   <desc>explicit mapping to an external LDAP dn for a given account</desc>
 </attr>
 
-<attr id="257" name="zimbraAuthFallbackToLocal" type="boolean" cardinality="single" optionalIn="globalConfig" flags="domainInherited">
+<attr id="257" name="zimbraAuthFallbackToLocal" type="boolean" cardinality="single" optionalIn="globalConfig,domain" flags="domainInherited">
   <globalConfigValue>TRUE</globalConfigValue>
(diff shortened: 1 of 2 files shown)


frisonisland (approved):

store/conf/attrs/attrs.xml
```diff
@@ -1470,7 +1470,8 @@ TODO - add support for multi-line values in globalConfigValue and defaultCOSValu
   <desc>explicit mapping to an external LDAP dn for a given account</desc>
 </attr>
 
-<attr id="257" name="zimbraAuthFallbackToLocal" type="boolean" cardinality="single" optionalIn="domain">
+<attr id="257" name="zimbraAuthFallbackToLocal" type="boolean" cardinality="single" optionalIn="globalConfig" flags="domainInherited">
```
keshavbhatt, line 1473: We also need to make it optionalIn Domain. 
So, it will go like this `optionalIn="globalConfig,domain"`.

keshavbhatt (changes requested):
Dropped a comment to fix the build issue.

Since the changes to attribute will also affect the globalCongig and Server entry I also suggest to run the code generator in store module `ant -Dcarbonio.buildinfo.version=22.8.0_ZEXTRAS_202208 generate-getters` in the store module to bring in the new changes.

store/conf/attrs/attrs.xml
keshavbhatt: @frisonisland It happens, no worries :)

Polpetta: Updated!

keshavbhatt (approved):

PR #96, opened 2022-09-20
feat: [CO-378] add proxy ip to log

frisonisland:
When OrigIp and OrigPort are not null or the remote address is localhost, remote address and port do not get logged.
In case of requests coming from of a trusted proxy the result is only the client ip and port get logged, but not the proxy server ip and port.

This PRs is to always add remote address and port, except when the remote address is localhost.


keshavbhatt (approved):
LGTM

Polpetta (approved):

PR #102, opened 2022-09-30
feat(CO-367): add carbonioMeshCredentials value

Polpetta:


Polpetta:
Before you ask, I dunno why all the spaces where trimmed, I just pasted my code at the end of the file

frisonisland (approved):

store/conf/attrs/attrs.xml
```diff
@@ -9840,4 +9840,8 @@ TODO: delete them permanently from here
   <globalConfigValue>TRUE</globalConfigValue>
   <desc>Whether Carbonio can take interactive feedback from users</desc>
 </attr>
+
+<attr id="3095" name="carbonioMeshCredentials" type="string" cardinality="single" optionalIn="globalConfig" since="22.11.0">
+  <desc>Carbonio base64 tarball credentials for Mesh service</desc>
```
keshavbhatt, line 9845: Let's make it more meaningful :)
```suggestion
  <desc>Carbonio mesh service credentials as base64 string</desc>
```

keshavbhatt (changes requested):
Suggested change in attribute description, and requested to run the java code generator to complete this addition.

Polpetta:
I've updated the generated Java classes!

keshavbhatt (approved):
To avoid code rearrangement on generated classes, you can skip running the Google Java Formatter (one of our pre-commit hooks). But everything should be fine even with that, just to avoid huge changes. I dont know why we dont get same output everytime with it.

frisonisland:
> To avoid code rearrangement on generated classes, you can skip running the Google Java Formatter (one of our pre-commit hooks). But everything should be fine even with that, just to avoid huge changes. I dont know why we dont get same output everytime with it.

I think it's time we think about not including generated classes in src and add some tests on them to check we get the default values we want ;)

keshavbhatt:
> > To avoid code rearrangement on generated classes, you can skip running the Google Java Formatter (one of our pre-commit hooks). But everything should be fine even with that, just to avoid huge changes. I dont know why we dont get same output everytime with it.
> 
> I think it's time we think about not including generated classes in src and add some tests on them to check we get the default values we want ;)

Yeah, in maven build we set the code generator to run everytime time the compile happen, i noticed the output remain same unless the attrs.xml is modified(which is good).
You mean tests for AttributeManager utility right?

frisonisland:
> > > To avoid code rearrangement on generated classes, you can skip running the Google Java Formatter (one of our pre-commit hooks). But everything should be fine even with that, just to avoid huge changes. I dont know why we dont get same output everytime with it.
> > 
> > 
> > I think it's time we think about not including generated classes in src and add some tests on them to check we get the default values we want ;)
> 
> Yeah, in maven build we set the code generator to run everytime time the compile happen, i noticed the output remain same unless the attrs.xml is modified(which is good). You mean tests for AttributeManager utility right?

Exactly! But I noticed the output does not match our lint rules, so each time it proposes me to push ZAttr classes, then runs the hook, fails, and then the files are formatted back as the origin.

keshavbhatt:
> I'm too for not committing generated code, I mean as long as it is exactly generated I don't see any reason to commit it +1 For now I'll merge as it is

Yes I agree, the reason why we need to commit it is beacuse it is part of the source code and required by other packages within mailbox, other projects to successfully build.

PR #106, opened 2022-10-07
fix:[CO-407] correctly check delegated account requests

keshavbhatt:
Aim: fix the OnBehalfOfRequest check.


frisonisland (approved):
Ok to me, but consider updating the javadoc if possible.
It doesn't make much sense from the description: if the request is delegated it's also on-behalf-of?

PR #111, opened 2022-10-26
chore: ACL rights generator

frisonisland:
### Goal
Add ACL rights management variables to the code

### What's changed

- Some missing ant targets responsible of code generation for ACL rights have been brought back. Now, by running generate-rights ant target, rights classes will be updated and file domainadmin-rights.xml will contain rights for domainAdminModifiable attributes (see attrs.xml)
- I have updated pre-commit hooks version "by accident" :slightly_smiling_face: (pre-commit autoupdate) but it should do no harm
- I have updated the build-common.xml because buildinfo.release version was using local username and it affected aut-generated files (release was my name). Now it defaults to carbonio.


store-conf/conf/msgs/ZsMsgRights.properties
```diff
@@ -1,13 +1,13 @@
 #
 #
 # DO NOT MODIFY - generated by RightManager
-# 8.7.0_BETA2_1501.zimbra zimbra 20160118-1614
+# unknown unknown unknown FOSS
```
keshavbhatt, line 4: Why RightManager did not picked the Carbonio version? I remember in past we delegated the zimbra buildinfo with carbonio buildinfo and this should also reflect the same, maybe we can also add the delgation logic to store build.xml?

store-conf/conf/rights/rights-domainadmin.xml
```diff
@@ -588,10 +644,7 @@
       <rights>
         <r n="accessGAL"/>
         <r n="checkDomainMXRecord"/>
-        <r n="countAccount" />
-        <r n="countAlias" />
-        <r n="countCalendarResource" />
-        <r n="countDistributionList" />
```
keshavbhatt, line 594: these appear to have been manually added by someone in the rights-domainadmin.xml file and are now being removed by autogenerated code, we must ensure that this does not impact the anticipated behaviour.

keshavbhatt (approved):
Added few concerns as comments, everything else look good to me.

store-conf/conf/msgs/ZsMsgRights.properties
frisonisland: I left it as is for the review, because I saw the version mismatch in first place and I was not sure it matters anymore (version 8.7.0 does not mean anything in our case). I'm not sure where it pickups the value from

frisonisland: I will check this again, because all the ZsMsgRights.properties are supposed to be auto-generated.

aheeva-yuliya (approved):

store-conf/conf/msgs/ZsMsgRights.properties
frisonisland: @keshavbhatt I fixed this file, it requires first building classes to generate BuildInfo, then calling generate-rights, then build again to include new generated Rights classes. Let me know if it is enough and want to close this conversation :slightly_smiling_face:

PR #112, opened 2022-11-02
fix: [CO-404] prevent duplicate account quota usages in GetQuotaUsage API

keshavbhatt:


store/src/java/com/zimbra/cs/service/admin/GetQuotaUsage.java
```diff
@@ -35,331 +23,476 @@
...
+   * @since 22.12.0
+   */
+  public static List<AccountQuota> getUniqueAccountQuotaList(List<AccountQuota> accountQuotas) {
+    return accountQuotas.stream()
+        .collect(
+            Collectors.collectingAndThen(
+                Collectors.toCollection(
+                    () -> new TreeSet<>(Comparator.comparing(AccountQuota::getId))),
+                ArrayList::new));
+  }
```
frisonisland, line 73: Hey @keshavbhatt I took some time testing this method and comparing TreeSet vs Using an HashMap or set to check for duplicates, because I was curious about TreeSet. I also took inspiration from here because there are some nice solutions: https://stackoverflow.com/questions/29670116/remove-duplicates-from-a-list-of-objects-based-on-property-in-java-8

I used 
```suggestion
  public static List<AccountQuota> getUniqueAccountQuotaList(List<AccountQuota> accountQuotas) {
    final Set<String> accountQuotaMap = new HashSet<>();
    return accountQuotas.stream().filter(a -> accountQuotaMap.add(a.getId())).collect(Collectors.toList());
  }
```
and found the following with increase data size (Took uses TreeSet, Set took uses HashSet):

500 accounts
Took: 75 ms.
Set Took: 2 ms.

5000 accounts
Took: 13 ms.
Set Took: 1 ms.

50000 accounts
Took: 15 ms.
Set Took: 15 ms.

5 million accounts
Took: 193 ms.
Set Took: 57 ms.

50 million accounts
Took: 1606 ms.
Set Took: 631 ms.

150 million accounts
Took: 4525 ms.
Set Took: 1659 ms.

It seems using HashSet comparison has an advantage, and I think it's because TreeSet is also sorting elements and that takes additional time (see its javadoc). Just something we should consider! Even with low numbers an API call would took 70ms less it seems.

frisonisland: Btw this is just to give an idea, it's just one run, would be interesting to see a sample with mean operation time, it just seems TreeSet computation time increases more. I made a chart and they both increase quadratic/exponentially it seems, but TreeSet just has a steeper curve.

frisonisland (approved):

aheeva-yuliya (approved):

PR #114, opened 2022-11-03
fix(appserver-service): depends service-discover + pending-setups

frisonisland:
Since the package uses pending-setups and service-discover it should depend on them.

In some cases this can lead to errors (was not able to replicate them, but was reported), for example if pending-setups was not installed the package will try to install some files on directories defined by pending-setups.

### DIFF
--- a/packages/appserver-service/PKGBUILD
+++ b/packages/appserver-service/PKGBUILD
@@ -15,6 +15,10 @@ url="https://zextras.com"
 optdepends=(
   "carbonio-common-appserver-docs"
 )
+depends=(
+  "service-discover"
+  "pending-setups"
+)
 
 section="mail"
 priority="optional"


aheeva-yuliya (approved):

Polpetta (approved):

packages/appserver-service/PKGBUILD
```diff
@@ -15,6 +15,10 @@ url="https://zextras.com"
 optdepends=(
   "carbonio-common-appserver-docs"
 )
+depends=(
+  "service-discover"
+  "pending-setups"
+)
```
keshavbhatt, line 21: This could be valid solution but since you are not able to replicate the issue. I personally will opt for another solution that guarantee to prevent any of package install order problem. In RPM world they deal with this issue with (https://docs.fedoraproject.org/en-US/packaging-guidelines/Scriptlets/#ordering). I found no way to define %triggerin in YAY(our package builder).

Following is the postinstall scriptlet of this package:

```
%post -p /bin/sh
getent group 'zextras' >/dev/null ||
groupadd -r 'zextras'
getent passwd 'zextras' >/dev/null ||
useradd -r -M -d '/opt/zextras' -g 'zextras' -s /bin/bash 'zextras'
getent group 'carbonio-mailbox' >/dev/null ||
groupadd -r 'carbonio-mailbox'
getent passwd 'carbonio-mailbox' >/dev/null ||
useradd -r -M -g 'carbonio-mailbox' -s /sbin/nologin 'carbonio-mailbox'
usermod -a -G 'carbonio-mailbox' 'zextras'
mkdir -p '/etc/zextras/carbonio-mailbox/'
chown 'carbonio-mailbox:carbonio-mailbox' '/etc/zextras/carbonio-mailbox/'
if [ -d /run/systemd/system ]; then
systemctl daemon-reload >/dev/null 2>&1 || :
systemctl enable carbonio-mailbox-sidecar.service >/dev/null 2>&1 || :
fi
echo "======================================================"
echo "Carbonio Mailbox Service installed successfully!"
echo "You must run pending-setups to configure it correctly."
echo "======================================================"
cat > /etc/zextras/pending-setups.d/carbonio-mailbox-setup.sh <<EOF
carbonio-mailbox setup
EOF
```
the scriptlet is failing because the directory `/etc/zextras/pending-setups.d/` doesn't exist.

My proposed solution would be: simply create the directory before creating the file(carbonio-mailbox-setup.sh) inside it. Feel free to take it or drop it if you are sure the existing solution will work.

keshavbhatt (approved):
Added another way to fix this issue, existing one also looks good to me.

packages/appserver-service/PKGBUILD
frisonisland: Thanks! I did it like that because I think pending-setups.d should be provided by pending-setups package. That would keep the configuration minimal, so each package manage its own resources.
Now if pending-setups is not being installed in the order defined by "depends" then we will have a problem, but I don't know why that would happen :neutral_face:

PR #115, opened 2022-11-07
feat: [CO-436] make proxyconfgen download domain certificates

Polpetta:
Upon `proxyconfgen` run, the certificates are now taken from LDAP and downloaded in `/opt/zextras/conf/domaincerts` but only if the domain has at least one `virtualHostname` set.


store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -328,11 +329,28 @@ private static String loadAllClientCertCA() {
...
+    if (!domainSSLDir.exists() && !domainSSLDir.mkdirs()) {
+      throw new ProxyConfException("Unable to create base domain certificate folder");
+    }
+    final File certificateFile =
+        new File(Path.of(DOMAIN_SSL_DIR, domainName + SSL_CRT_EXT).toUri());
+    final File privateKeyFile = new File(Path.of(DOMAIN_SSL_DIR, domainName + SSL_KEY_EXT).toUri());
+    if ((certificateFile.exists() && !certificateFile.delete())
+        || (privateKeyFile.exists() && !privateKeyFile.delete())) {
+      throw new ProxyConfException(
+          "Unable to remove broken certificates before downloading new ones from LDAP");
```
frisonisland, line 344: Consider switch "broken" for "old". I got that "broken" is because they cannot be removed, but it may seem also that certificates content is broken.

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -328,11 +329,28 @@ private static String loadAllClientCertCA() {
...
+    }
+    try (FileOutputStream fsOutputCertificate = new FileOutputStream(certificateFile);
+        FileOutputStream fsOutputPrivateKey = new FileOutputStream(privateKeyFile)) {
+      LOG.debug("Deploying certificate for " + domainName);
+      fsOutputCertificate.write(certificate.getBytes(StandardCharsets.UTF_8));
+      LOG.debug("Deploying private key for " + domainName);
+      fsOutputPrivateKey.write(privateKey.getBytes(StandardCharsets.UTF_8));
+    } catch (IOException e) {
+      throw new ProxyConfException("Unable to write down certificates for domain " + domainName, e);
```
frisonisland, line 353: What if we delete the files and the write fails? 
Do you think doing a backup of the files (e.g.:with .bkp extension) , writing new one, then deleting or restoring the backup, based on operation success, would be a better idea?

frisonisland (changes requested):
I just left a couple of comments, I would like to check other options in case of process failure before proceeding!

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -328,11 +329,28 @@ private static String loadAllClientCertCA() {
...
-      //noinspection ResultOfMethodCallIgnored
-      domainSSLDir.mkdirs();
+    if (!domainSSLDir.exists() && !domainSSLDir.mkdirs()) {
+      throw new ProxyConfException("Unable to create base domain certificate folder");
+    }
+    final File certificateFile =
+        new File(Path.of(DOMAIN_SSL_DIR, domainName + SSL_CRT_EXT).toUri());
+    final File privateKeyFile = new File(Path.of(DOMAIN_SSL_DIR, domainName + SSL_KEY_EXT).toUri());
+    if ((certificateFile.exists() && !certificateFile.delete())
+        || (privateKeyFile.exists() && !privateKeyFile.delete())) {
```
keshavbhatt, line 342: consider using java.nio.file.Files#delete for deletion as it better indicates the cause of the failure in exception.

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
keshavbhatt: I agree, It would be preferable if the entire change operation could be reversed in the case of any failure.

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -328,11 +329,28 @@ private static String loadAllClientCertCA() {
     return sb.toString();
   }
 
-  public static void createDomainSSLDirIfNotExists() {
+  public static void downloadDomainCertificate(
```
keshavbhatt, line 332: Add javadoc.

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -251,12 +248,19 @@ private static List<DomainAttrItem> loadDomainReverseProxyAttrs() throws Service
...
-                  && privateKey == null
-                  && clientCertMode == null
-                  && clientCertCA == null)) {
+          if (virtualHostnames.length != 0
+              && !ProxyConfUtil.isEmptyString(certificate)
+              && !ProxyConfUtil.isEmptyString(privateKey)) {
+            // download the certificates from LDAP if there is at least a virtual host name for that
+            // domain
+            try {
+              downloadDomainCertificate(domainName, certificate, privateKey);
```
keshavbhatt, line 257: Can we move this out of this method? 
**why:** this method `List<DomainAttrItem> loadDomainReverseProxyAttrs()` by definition looks like should just return list of attributes and do nothing.

keshavbhatt (changes requested):

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
Polpetta: I mean, they already decided to overwrite them when uploading new certificates in the admin UI interface, so I think this backup thingy should be done at LDAP level, not when regenerating the nginx configuration...it's already too late here imho

keshavbhatt:
Also add a dry-run check conditional to prevent this from being executed with the -n option.

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -350,7 +353,13 @@ public static void downloadDomainCertificate(
       LOG.debug("Deploying private key for " + domainName);
       fsOutputPrivateKey.write(privateKey.getBytes(StandardCharsets.UTF_8));
     } catch (IOException e) {
-      throw new ProxyConfException("Unable to write down certificates for domain " + domainName, e);
+      throw new ProxyConfException(
```
frisonisland, line 356: @Polpetta sorry, I think we have to move the certificates back :sweat_smile: otherwise Nginx won't be able to load them!
I would apply this logic:

- if ok remove backup files
- if fails put them back

frisonisland (changes requested):

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
Polpetta: I'd relly avoid doing the rollback otherwise the codebase will be a mess :(

Polpetta: I'd have to do stuff inside a `catch` block :(

Polpetta: And the operation could still fail if, for instance, the machine has run out of space

frisonisland: Understood...what about just writing new content to files instead of deleting them? We'll also have less code. 
See the append option: https://docs.oracle.com/javase/8/docs/api/java/io/FileOutputStream.html#FileOutputStream-java.io.File-boolean-
The file permission thing is a corner case that should not happen anyway in a normal system.

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
Polpetta: I used `Files.move` to backup the previous file

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
Polpetta: Done!

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
Polpetta: This is now outdated

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
Polpetta: I did it anyway I feels sorry for the next dude that will see another try inside a catch block :D

Polpetta:
@keshavbhatt @frisonisland we updated the PR please check again

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
frisonisland: It will be probably me :D not sure if vavr would have helped

frisonisland (approved):

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -328,11 +310,65 @@ private static String loadAllClientCertCA() {
...
   }
 
-  public static void createDomainSSLDirIfNotExists() {
-    File domainSSLDir = new File(DOMAIN_SSL_DIR);
-    if (!domainSSLDir.exists()) {
-      //noinspection ResultOfMethodCallIgnored
-      domainSSLDir.mkdirs();
+  /**
+   * This method will write down the certificate and the private key for a provided domain. It
+   * creates a backup of the old one
```
keshavbhatt, line 315:
```suggestion
   * This method backs up existing certificate and private key files for the given domain before creating
   * new ones and populating them with the new data. Backup files gets restored if the transaction fails.
```

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -328,11 +310,65 @@ private static String loadAllClientCertCA() {
...
+   *
+   * @param domainName the name of the domain which the certificate and the private key will have
+   *     the name
+   * @param certificate the certificate content
+   * @param privateKey the private key content
+   * @throws ProxyConfException if something goes wrong :)
```
keshavbhatt, line 321: remove emoji :D

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -328,11 +310,65 @@ private static String loadAllClientCertCA() {
...
+   * @author Davide Polonio and Yuliya Aheeva
+   */
+  public static void writeDownCertificate(String domainName, String certificate, String privateKey)
```
keshavbhatt, line 324: consider naming this method `updateDomainCertificate`

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -2359,6 +2400,35 @@ public static int createConf(String[] args) throws ServiceException, ProxyConfEx
...
+  }
+
+  /**
+   * Creates a folder
+   *
+   * @param folderPath the path to create
+   * @throws ProxyConfException if something goes wrong
+   * @author Davide Polonio and Yuliya Aheeva
+   */
+  private static void createFolder(String folderPath) throws ProxyConfException {
```
keshavbhatt, line 2425: if you think this utility method is too generic move it to `com.zimbra.cs.util.proxyconfgen.Utils`

keshavbhatt (changes requested):
added few suggestion, everything else looks good to me. 
unit tests are still missing.

keshavbhatt (approved):
LGTM

PR #124, opened 2022-11-22
fix: [CO-464] upload cert and private key only if they have value

aheeva-yuliya:


keshavbhatt (approved):
lgtm

Polpetta (approved):

frisonisland (approved):
Ok, makes sense to write them if they are not both empty

frisonisland (changes requested):
If this is an hotfix let's reopen it from main

PR #126, opened 2022-11-24
fix:[CO-457] delete orphaned cert-key pair from domain SSL dir

keshavbhatt:
**What has changed:**
- The domain SSL directory is cleaned to remove orphaned certificate key pair.

**Test:** Changes are manually tested, its really hard to write unit test for this class as
methods and members are static everywhere.

### DIFF
--- a/store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
+++ b/store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
@@ -2436,7 +2436,14 @@ public class ProxyConfGen {
                   "Will delete obsoleted domain %s file %s",
                   (fileName.endsWith(SSL_KEY_EXT) ? "key" : "cert"), fileName));
     } else {
-      filesInDirectory.forEach(Utils::deleteFileIfExists);
+      filesInDirectory.forEach(
+          filePath -> {
+            try {
+              Utils.deleteFileIfExists(filePath);
+            } catch (ProxyConfException e) {
+              LOG.info(e.getMessage());
+            }
+          });
     }
   }
 
--- a/store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
+++ b/store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
@@ -327,11 +327,6 @@ public class ProxyConfGen {
    */
   public static void updateDomainCertificate(
       String domainName, String certificate, String privateKey) throws ProxyConfException {
-
-    if (certificate == null || privateKey == null) {
-      return;
-    }
-
     final File certificateFile =
         new File(Path.of(DOMAIN_SSL_DIR, domainName + SSL_CRT_EXT).toUri());
     final File privateKeyFile = new File(Path.of(DOMAIN_SSL_DIR, domainName + SSL_KEY_EXT).toUri());
--- a/store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
+++ b/store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
@@ -36,6 +36,7 @@ import org.apache.commons.cli.Options;
 import org.apache.commons.cli.ParseException;
 
 public class ProxyConfGen {
+
   static final String ZIMBRA_USER = "zextras";
   static final String ZIMBRA_UPSTREAM_NAME = "zimbra";
   static final String ZIMBRA_UPSTREAM_WEBCLIENT_NAME = "zimbra_webclient";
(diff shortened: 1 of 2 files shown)


store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -2401,6 +2414,32 @@ public static int createConf(String[] args) throws ServiceException, ProxyConfEx
...
+   *     from domains
+   * @param dryRun if it's a dry run
+   * @author Keshav Bhatt
+   * @since 22.12.0
+   */
+  private static void deleteObsoleteCertificateKeyPair(
+      final List<DomainAttrItem> mDomainReverseProxyAttrs, final boolean dryRun) {
+    List<String> filesInDirectory = Utils.getFilesPathInDirectory(DOMAIN_SSL_DIR);
+    for (DomainAttrItem entry : mDomainReverseProxyAttrs) {
+      filesInDirectory.removeIf(filePath -> (filePath.contains(entry.domainName)));
```
Polpetta, line 2430: Sorry but doesn't this delete a certificate that currently exists? Souldn't it be
```suggestion
      filesInDirectory.removeIf(filePath -> !(filePath.contains(entry.domainName)));
```
? Note that the certificate is moved from `certificate.crt` to `certificate.crt.bak` before every run, overwriting the old `bak` one.

store/src/java/com/zimbra/cs/util/proxyconfgen/Utils.java
```diff
@@ -49,4 +54,39 @@ public static void createFolder(String folderPath) throws ProxyConfException {
...
+   * @author Keshav Bhatt
+   * @since 22.12.0
+   */
+  public static void deleteFileIfExists(final String filePath) {
+    File file = new File(filePath);
+    if (file.exists()) {
+      try {
+        Files.delete(file.toPath());
+      } catch (final IOException ignored) {
+        // ignore
```
Polpetta, line 71: Are you sure we want to ignore if the deletion goes wrong? Not even a warning to the final user?

Polpetta (changes requested):
I've a question for you :grin:

store/src/java/com/zimbra/cs/util/proxyconfgen/Utils.java
keshavbhatt: I avoided it since it will cause confussions for user in this specific case. But I will add since you asked.

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
Polpetta: Oh yeah right, I'm dumb :sweat_smile:

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -2401,6 +2414,39 @@ public static int createConf(String[] args) throws ServiceException, ProxyConfEx
...
+      filesInDirectory.forEach(
+          fileName ->
+              LOG.info(
+                  "Will delete obsoleted domain %s file %s",
+                  (fileName.endsWith(SSL_KEY_EXT) ? "key" : "cert"), fileName));
+    } else {
+      filesInDirectory.forEach(
+          filePath -> {
+            try {
+              Utils.deleteFileIfExists(filePath);
```
Polpetta, line 2442: Please check if there are old `bak` files for `crt` and `key` too

Polpetta (approved):
:+1: don't forget the `bak` files too!

store/src/java/com/zimbra/cs/util/proxyconfgen/Utils.java
Polpetta: It makes sense

store/src/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
keshavbhatt: They are already covered.

frisonisland: Btw I know that when we generate the config we leave the old .bak files, but my idea with .bak files was having them just during the generation process in case writing new file fails, if not remove them.
Let me explain:
- upload certificates to ldap
- regenerate configuration
  - make .bak files of certificates
  - write certificates
    - if all ok remove .bak
    - if fails put back .bak files
- restart nginx

If certificates are invalid for nginx for some reason then the user should put back in ldap old values and not manually put .bak files as original. We can create a new task if you agree.

aheeva-yuliya (approved):

PR #131, opened 2022-11-30
chore: maven build

frisonisland:
This is the new PR for maven build changes. Reason is old branch had files moved around and they lost history.
Another benefit is that here there are fewer commits, so it may be easier for you to see the changes.
Unfortunately lot of them address test resources, because I still had some troubles with files line separator, etc.

### Goal
Moving the project to maven.
This will allow us to fix many dependencies issue we have, and also be able to add new features more easily with all the tools we need (e.g.: TestContainers, etc.)

### Memorable changes
- **soap-doc module** has been removed and the code is now **inside soap**
- **store module is a WAR**
- **dependency management** declared in parent **zm-mailbox pom**
- **store compiles twice**: pre-compile, generate-sources, compile again
- **carbonio.buildinfo.version statically declared in ant-build-common.xml**


Polpetta (approved):
I've reviewed everything :new_moon_with_face: , of course LGTM

keshavbhatt (approved):
LGTM

PR #138, opened 2023-01-09
feat: [CO-477] Set default for 'catch-all messages' address in alias domains

keshavbhatt:
**What has changed:**
- Provisioning now sets mailCatchAllForwardingAddress domain attribute to default value during alias domain creation.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -27,3 +27,6 @@ build.properties
 build
 **/**/out/
 .flattened-pom.xml
+
+target/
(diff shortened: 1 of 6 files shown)


Polpetta (approved):

frisonisland (approved):
LGTM, if I got it right when we create an alias domain we set the original domain name in the "catchAll" attribute.

aheeva-yuliya (approved):

PR #140, opened 2023-01-10
chore: Remove build info from disclaimer header

keshavbhatt:
**What has changed:**
- remove build info from disclaimer header of generated files. This was done to prevent unnecessary file changes caused by dynamic build number. This information if required can be retrieved from package manager


keshavbhatt:
> It's ok to me as long as we are aware that all LDAP schema files will not include version in the header after these changes!

I will test the impact tomorrow, by spinning a full build. I have a strong feeling that it wont break :face_in_clouds:

PR #144, opened 2023-01-24
chore: soap docs style & buildinfo for soap module

keshavbhatt:
**What has changed:**

~~-**fix: generate, package buildinfo class**~~
    ~~- Now generate and package buildinfo class for soap module in respective mvn phases.~~ (NOT REQUIRED, using version property defined by ant task)
     
- **fix: set build version from ant-build-common's target**
  - This also accomodate the soap docs generator to have correct build info version in the generated files

- **fix: properly set anchor scroll offset**
    -  Now offsets html anchors according to the height of fixed header

### DIFF
--- a/soap/ant-generate-soap-docs.xml
+++ b/soap/ant-generate-soap-docs.xml
@@ -70,7 +70,7 @@
       public static final String HOST = "${zimbra.buildinfo.host}";
       }
     </echo>
-    <javac includeantruntime="false" destdir="${build.classes.dir}" debug="true"
+    <javac includeantruntime="false" destdir="${build.classes.dir}"
       target="${javac.target}" srcdir="${maven.build.dir}/buildinfo">
       <compilerarg value="-Xlint:deprecation"/>
       <compilerarg value="-Xlint:unchecked"/>
--- a/soap/ant-generate-soap-docs.xml
+++ b/soap/ant-generate-soap-docs.xml
@@ -21,7 +21,7 @@
     <pathelement path="${test_classpath}"/>
   </path>
 
-  <target name="check-schema-done">
+  <target name="check-schema-done" depends="set-dev-version">
     <uptodate property="schema.build.notrequired" targetFile="${xml.schema.dir}/zimbra.xsd">
       <srcfiles dir="${src.dir}" includes="**/*.java" />
     </uptodate>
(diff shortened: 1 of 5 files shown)


soap/ant-generate-soap-docs.xml
```diff
@@ -48,6 +49,34 @@
...
+      public static final String RELCLASS = "${zimbra.buildinfo.relclass}";
+      public static final String RELNUM = "${zimbra.buildinfo.relnum}";
+      public static final String BUILDNUM = "${zimbra.buildinfo.buildnum}";
+      public static final String VERSION = "${zimbra.buildinfo.version}";
+      public static final String RELEASE = "${zimbra.buildinfo.release}";
+      public static final String DATE = "${DSTAMP}-${TSTAMP}";
+      public static final String HOST = "${zimbra.buildinfo.host}";
+      }
+    </echo>
+    <javac includeantruntime="false" destdir="${build.classes.dir}" debug="true"
```
Polpetta, line 73: `debug="true"` are you sure about it?

Polpetta (approved):
I think it is fine. I left a comment just to clarify a doubt

soap/ant-generate-soap-docs.xml
keshavbhatt: Thanks for pointing this out. I think we can get rid of it, as it is set to "off" by default

 **according to:**
https://ant.apache.org/manual/Tasks/javac.html

> debug
> Indicates whether source should be compiled with debug information. If set to off, -g:none will be passed on the > command line for compilers that support it (for other compilers, no command line argument will be used). If set > to true, the value of the debuglevel attribute determines the command line argument.

aheeva-yuliya (approved):

soap/pom.xml
```diff
@@ -172,9 +172,39 @@
...
           </execution>
+          <!-- generate sources -->
+          <execution>
+            <configuration>
+              <target>
+                <property name="maven_compile_classpath" refid="maven.compile.classpath"/>
+                <taskdef classpathref="maven.plugin.classpath"
+                  resource="net/sf/antcontrib/antlib.xml"/>
+                <ant antfile="${basedir}/ant-generate-soap-docs.xml">
+                  <target name="generate-sources"/>
```
frisonisland, line 183: This execution will run generate-sources ant target. I suggest just executing the specific task generate-buildinfo.

soap/ant-generate-soap-docs.xml
```diff
@@ -48,6 +49,34 @@
     </java>
   </target>
 
+  <target name="generate-sources" depends="generate-buildinfo"/>
```
frisonisland, line 52: Why define a task called generate-sources that only calls another task?
To keep low maintenance I suggest using just generate-buildinfo or we will end up having too many ant task that are difficult to track down.
Or "generate-sources" could make sense if it collects multiple task for example.
What do you say?

frisonisland (changes requested):

soap/pom.xml
keshavbhatt: My idea was to wrap all sourece code generation releated tasks under this. Easy to look for from maven phase pov.

soap/ant-generate-soap-docs.xml
keshavbhatt: I agree with that, the idea was to also put other soruce code generation readted task depend on this, so from maven side we set biding for generate-sources ant task :)

frisonisland: Ok, keep in mind that maven has multiple generate-source task anyway, with different ids.
I agree with the idea of defining on "generate-sources" on ant tasks, maybe let's see if we can collect the generate-buildinfo and generate-soap-api in one task?

frisonisland (approved):
Ok to me! Consider updating the PR description if we don't need the BuildInfoGenerated class!

PR #145, opened 2023-01-27
feat: [CO-526] prevent delegated admins creating DDLs

keshavbhatt:
**What has changed:**
- prohibit delegatedAdmins from creating dynamic distribution list using CreateDistributionList SOAP command
    - Now throws ServiceException when a delegatedAdmin requests to create dynamic distribution list

### DIFF
--- a/store/src/test/java/com/zimbra/cs/service/admin/CreateDistributionListTest.java
+++ b/store/src/test/java/com/zimbra/cs/service/admin/CreateDistributionListTest.java
@@ -1,0 +1,112 @@
+package com.zimbra.cs.service.admin;
+
+import static org.mockito.ArgumentMatchers.anyBoolean;
+import static org.mockito.ArgumentMatchers.anyMap;
+import static org.mockito.ArgumentMatchers.anyString;
+import static org.mockito.Mockito.doReturn;
+import static org.mockito.Mockito.mock;
+import static org.mockito.Mockito.spy;
+import static org.mockito.Mockito.when;
+
+import com.zimbra.common.account.ProvisioningConstants;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.soap.SoapProtocol;
+import com.zimbra.cs.account.Account;
+import com.zimbra.cs.account.AuthToken;
+import com.zimbra.cs.account.DynamicGroup;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.cs.mailbox.MailboxTestUtil;
+import com.zimbra.soap.JaxbUtil;
+import com.zimbra.soap.SoapEngine;
+import com.zimbra.soap.ZimbraSoapContext;
+import com.zimbra.soap.admin.message.CreateDistributionListRequest;
+import com.zimbra.soap.admin.type.Attr;
+import java.util.ArrayList;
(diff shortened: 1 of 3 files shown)


store/src/test/java/com/zimbra/cs/service/admin/CreateDistributionListTest.java
```diff
@@ -0,0 +1,112 @@
...
+    //create CreateDistributionListRequest
+    List<Attr> attributes = new ArrayList<>();
+    attributes.add(new Attr(Provisioning.A_memberURL,
+        "ldap:///??sub?(&amp;(objectClass=zimbraAccount)(ZimbraAccountStatus=active))"));
+    attributes.add(new Attr(Provisioning.A_zimbraIsACLGroup, ProvisioningConstants.TRUE));
+    final CreateDistributionListRequest createDistributionListRequest = new CreateDistributionListRequest(
+        DISTRIBUTION_LIST_NAME, attributes, true);
+
+    //create CreateDistributionList spy
+    final CreateDistributionList createDistributions = spy(CreateDistributionList.class);
```
Polpetta, line 98: You can use a real object here, no need to spy

Polpetta (changes requested):
Check the comment I left :call_me_hand:

store/src/test/java/com/zimbra/cs/service/admin/CreateDistributionListTest.java
keshavbhatt: I agree. the issue is: with real object it fails causing issues on `checkDomainRightByEmail` inside handle.

store/src/test/java/com/zimbra/cs/service/admin/CreateDistributionListTest.java
```diff
@@ -0,0 +1,112 @@
...
+
+public class CreateDistributionListTest {
+
+  private final static String DOMAIN_NAME = "test.com";
+  private final static String ADMIN_USER_NAME = "admin";
+  private final static String DOMAIN_ADMIN_EMAIL = ADMIN_USER_NAME + "@" + DOMAIN_NAME;
+  private final static  String DISTRIBUTION_LIST_NAME = "developers" + "@" + DOMAIN_NAME;
+  private final static String DOMAIN_ADMIN_PASSWORD = "assext";
+
+  private static Provisioning provisioningSpy;
```
frisonisland, line 42: What about not spying the Provisioning? I remember we achieved the result without needing to spy in old tests

frisonisland (changes requested):

frisonisland (approved):
Sorry, no need to block for a spy, just check if it can be done by calling the real Provisioning

keshavbhatt:
> Sorry, no need to block for a spy, just check if it can be done by calling the real Provisioning

@frisonisland I already checked :)  Provisioning::createGroup method involved in the handler requires real provisioning backend, since its a unit test i skipped setting up full fledged provisioning backend. I think spy is the best way to achieve this since only this methos is stubbed whole provisioning is intact and used whenever required.

aheeva-yuliya (approved):

Polpetta (approved):

PR #150, opened 2023-02-03
fix: [CO-544] modify domain remove vhost names

frisonisland:
## Goal

When using zmprov we want to remove all virtual hostnames on a domain by executing:
> zmprov md <domain_name> zimbraVirtualHostname ""
> 

and not having to remove them all by specifying each one.

## What's changed
This PR adds the fix by matching the expected behavior.
If zimbraVirtualHostname are not passed then existing values will not be overridden, **the condition for removal is met when you pass exactly a zimbraVirtualHostname array containing only the empty string.**

### About MockProvisioning
I found MockProvisioning has many limitations as it does not reflect the real underlying behavior of the production system in some cases.
For example removing an attribute by setting it to empty value did not work (it should AFAIK), so I changed the class to match real-scenario.

## Tests
Code aside, **tested on a VM using zmprov.**


keshavbhatt (approved):

aheeva-yuliya (approved):

Polpetta (approved):
lgtm

PR #153, opened 2023-02-09
fix: [CO-491] caldav schedule agent client and none

frisonisland:
### Goal
When the client sends a CalDAV request with SCHEDULE-AGENT=CLIENT, the server should not send a message for the meeting.

### What's changed

* Added support for SCHEDULE-AGENT=CLIENT (no scheduling)
* Added support for SCHEDULE-AGENT=NONE (no scheduling)

Please check changes locally because these java files are big and linting changed everything.
But:
* I have added SCHEDULE_AGENT on ZCalendar.ICalTok
* I have added ZCalendar.ScheduleAgent Enum
* Changed the Invite class to set scheduleAgent field on ZOrganizer and ZAttendee when receiving the Calendar request
* Modified OrganizerAutoScheduler to return email address of attendees that do not have SCHEDULE-AGENT=CLIENT or NONE (these should not be scheduled as message)
* Modified CalendarUser to support scheduleAgent field (ZOrganizer and ZAttendee are children of it)

### Tests
* See OrganizerAutoScheduleTest
* Tested with Outlook CDS: when SCHEDULE-AGENT=CLIENT Carbonio does not send invitation email, only Outlook.

### DIFF
--- a/common/src/main/java/com/zimbra/common/calendar/ZCalendar.java
+++ b/common/src/main/java/com/zimbra/common/calendar/ZCalendar.java
@@ -20,6 +20,7 @@ import java.io.Writer;
 import java.util.ArrayList;
 import java.util.Iterator;
 import java.util.List;
+import java.util.Objects;
 import java.util.regex.Pattern;
 import net.fortuna.ical4j.data.CalendarParser;
 import net.fortuna.ical4j.data.CalendarParserImpl;
@@ -38,10 +39,32 @@ public class ZCalendar {
    * @author davidefrison
    * @since 23.3.0
    */
-  public static enum ScheduleAgent {
+  public enum ScheduleAgent {
     SERVER,
     CLIENT,
-    NONE
+    NONE;
+
+    /**
+     * Returns a {@link ScheduleAgent} from a string value by doing a case-insensitive match. It
+     * returns null in case of no matches.
+     *
+     * @param scheduleAgentString input string value of schedule agent
+     * @return a schedule agent
+     */
+    public static ScheduleAgent getScheduleAgent(String scheduleAgentString) {
+      if (!(Objects.isNull(scheduleAgentString))) {
+        final String scheduleAgentLower = scheduleAgentString.toLowerCase();
+        switch (scheduleAgentLower) {
+          case "server":
+            return ScheduleAgent.SERVER;
+          case "client":
+            return ScheduleAgent.CLIENT;
+          case "none":
+            return ScheduleAgent.NONE;
+        }
+      }
+      return null;
+    }
   }
 
   public static enum ICalTok {
--- a/common/src/main/java/com/zimbra/common/calendar/ZCalendar.java
+++ b/common/src/main/java/com/zimbra/common/calendar/ZCalendar.java
@@ -32,6 +32,18 @@ public class ZCalendar {
   public static final String sIcalVersion = "2.0";
   public static final String sObsoleteVcalVersion = "1.0";
 
+  /**
+   * Values for SCHEDULE-AGENT param (https://www.rfc-editor.org/rfc/rfc6638.html#section-7.1)
+   *
+   * @author davidefrison
+   * @since 23.3.0
+   */
+  public static enum ScheduleAgent {
+    SERVER,
+    CLIENT,
+    NONE
+  }
+
   public static enum ICalTok {
     ACTION,
     ALTREP,
--- a/common/src/main/java/com/zimbra/common/calendar/ZCalendar.java
+++ b/common/src/main/java/com/zimbra/common/calendar/ZCalendar.java
@@ -5,6 +5,11 @@
 
 package com.zimbra.common.calendar;
 
+import com.google.common.collect.ImmutableList;
+import com.zimbra.common.localconfig.LC;
+import com.zimbra.common.mime.MimeConstants;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.common.util.ZimbraLog;
 import java.io.BufferedInputStream;
 import java.io.ByteArrayInputStream;
 import java.io.IOException;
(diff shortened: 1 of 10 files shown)


common/src/main/java/com/zimbra/common/calendar/ZCalendar.java
```diff
@@ -16,997 +21,1209 @@
...
+  public static final String sIcalVersion = "2.0";
+  public static final String sObsoleteVcalVersion = "1.0";
+
+  /**
+   * Values for SCHEDULE-AGENT param (https://www.rfc-editor.org/rfc/rfc6638.html#section-7.1)
+   *
+   * @author davidefrison
+   * @since 23.3.0
+   */
+  public static enum ScheduleAgent {
```
keshavbhatt, line 41: remove static keyword, its static by default.

store/src/main/java/com/zimbra/cs/dav/caldav/AutoScheduler.java
```diff
@@ -42,573 +34,755 @@
...
+     * @param attendees meeting attendees
+     * @return list of email addresses
+     */
+    public List<Address> getRecipientsForAttendees(List<ZAttendee> attendees) {
+      if (attendees == null || attendees.isEmpty()) {
+        return Lists.newArrayListWithCapacity(0);
+      }
+      List<Address> recipients = Lists.newArrayListWithCapacity(attendees.size());
+      for (ZAttendee attendee : attendees) {
+        if (Objects.equals(attendee.getScheduleAgent(), "CLIENT")
```
keshavbhatt, line 622: instead of hardcoding CLIENT and NONE (which quickly makes no sense) consider using already defined constants in ScheduleAgent enum.

frisonisland: Thanks I added them for that very purpose, I forgot to use them :sweat_smile:

store/src/main/java/com/zimbra/cs/mailbox/calendar/CalendarUser.java
```diff
@@ -5,312 +5,343 @@
...
+  private String mCn;
+  private String mSentBy;
+  private String mDir;
+  private String mLanguage;
+
+  public String getScheduleAgent() {
+    return scheduleAgent;
+  }
+
+  public CalendarUser setScheduleAgent(String scheduleAgent) {
```
keshavbhatt, line 44: consider using already defined ScheduleAgent enum as method parameter instead of string.

keshavbhatt (changes requested):
I requested few changes to improve code quality.
Everything else looks good.

store/src/main/java/com/zimbra/cs/mailbox/calendar/CalendarUser.java
frisonisland: Ok, the initial idea was just to use it to check values, as the schedule agent on CalendarUser is what we get from the request (if nothing fallbacks to "SERVER") and there could be corner cases like wrong case, etc. so by changing this we should take extra care of parsing the input value. I will change it.

frisonisland: I'm doing some changes, there will be more boilerplate basically

frisonisland:
Reading https://www.rfc-editor.org/rfc/rfc5545 (https://www.rfc-editor.org/rfc/rfc6638.html extends it) property parameters value are intended to be case-insensitive but this is not covered yet, I will add changes.

store/src/main/java/com/zimbra/cs/mailbox/calendar/CalendarUser.java
keshavbhatt: > Ok, the initial idea was just to use it to check values, as the schedule agent on CalendarUser is what we get from the request (if nothing fallbacks to "SERVER") and there could be corner cases like wrong case, etc. so by changing this we should take extra care of parsing the input value. For example the value "serVer" cannot be translated to ScheduleAgent, same for null string, do we still want to proceed?
> 
> The alternative is to just accept whatever comes from the client, if it does not match "CLIENT" or "NONE" then it sends the message.

I understand that using the string param will make it flexible but the idea is to follow the RFC. The RFC defines strict rules and conditions that must be followed. And to enforce them we must consider restricting to use constants defined in the ENUM. 

I think it will be simple to handle exceptional cases:
if SCHEDULE-AGENT param does not matches the strictly defined values, we can assume the parameter got wrong value and in that case user SERVER as suggested in RFC. I quote from (https://www.rfc-editor.org/rfc/rfc6638.html#section-7.1 [Page 42]):

>This property parameter MAY be specified on "ORGANIZER"
      or "ATTENDEE" iCalendar properties.  In the absence of this
      parameter, the value "SERVER" MUST be used for the default
      behavior.

frisonisland: Agree, also since it is case-insensitive it makes more sense at this point to map the value to a state of Schedule-Agent to avoid issues.

store/src/main/java/com/zimbra/cs/mailbox/calendar/CalendarUser.java
```diff
@@ -37,16 +39,37 @@ public abstract class CalendarUser {
...
-    return this;
   }
 
-  private String scheduleAgent;
+  /**
+   * Converts a schedule agent value from string
+   *
+   * @param scheduleAgent
+   */
+  public void setScheduleAgentFromString(String scheduleAgent) {
```
keshavbhatt, line 55: Why this is required?

frisonisland: Because we need to translate the incoming request to ScheduleAgent for ZOrganzier and ZAttendee.
I could have done it in a private method of Invite.java class, but that would have meant to implement such method each time converting case-insensitive pops up, so I thought better add it to the enum.

frisonisland: Plus using .valueOf() parses only case-sensitive inputs and throws exception if provided value is not in enums so it does not cover the scenario.

keshavbhatt: What if this method gets arbitary string that is not covered by the enum? I dont see no default case. :confused:

frisonisland: It returns null at the end! I can add the default, np

keshavbhatt: This is throwing IllegalArgumentException. Please add a unit test for this method to confirm possible edge cases.

keshavbhatt (changes requested):

keshavbhatt (approved):
lgtm.

Polpetta (approved):

frisonisland:
I won't make other code changes since everyone approved, but will add some new tests to Invite class since it was never covered.

PR #156, opened 2023-02-20
feat: [CO-539] Add attributes to manage infrastructure notifications

keshavbhatt:
**What has changed:**
- Following attributes were added:
    - carbonioNotificationFrom
    - carbonioNotificationRecipients

**Related PRs:**
-


store/conf/attrs/attrs.xml
```diff
@@ -9989,4 +9989,12 @@ TODO: delete them permanently from here
   <desc>Logo URL for domain</desc>
   <globalConfigValue>https://www.zextras.com</globalConfigValue>
 </attr>
+
+<attr id="3127" name="carbonioNotificationFrom" type="email" max="256" cardinality="single" optionalIn="globalConfig,domain" flags="domainAdminModifiable" since="23.3.0">
```
Polpetta, line 9993: The maximum length of an email address is 254 characters, not 256! See https://www.rfc-editor.org/errata/eid1690
```suggestion
<attr id="3127" name="carbonioNotificationFrom" type="email" max="254" cardinality="single" optionalIn="globalConfig,domain" flags="domainAdminModifiable" since="23.3.0">
```

store/conf/attrs/attrs.xml
```diff
@@ -9989,4 +9989,12 @@ TODO: delete them permanently from here
...
+  <desc>email address of sender used to send emails about important infrastructure notifications</desc>
+</attr>
+
+<attr id="3128" name="carbonioNotificationRecipients" type="email" max="256" cardinality="multi" optionalIn="globalConfig,domain" flags="domainAdminModifiable" since="23.3.0">
```
Polpetta, line 9997: Same
```suggestion
<attr id="3128" name="carbonioNotificationRecipients" type="email" max="254" cardinality="multi" optionalIn="globalConfig,domain" flags="domainAdminModifiable" since="23.3.0">
```

Polpetta (changes requested):
There's just the email max length to adjust, but the rest is :fire:

store/conf/attrs/attrs.xml
keshavbhatt: @Polpetta  I have no issue accepting proposed changes, I was wondering why everywhere in mailbox code we are using 256 as max limit. I was trying to follow the same pattern. But the reasoning behind using 256 max limit seems okay according to RFC for SMTP https://www.rfc-editor.org/rfc/rfc5321#section-4.5.3

frisonisland: I was reading all these RFCs and it seems:
 - 64 chars for user part + 255 chars for domain part + "@" = 320 chars total (original RFC: https://www.rfc-editor.org/rfc/rfc3696)
 - then they said in errata the total of the path should be 256 (https://www.ietf.org/rfc/rfc2821.txt)
 - Since path contains the brackets "<" and ">" then total for user@domain is limited to 254

The RFC for SMTP is talking about the full path, hence the value 256.
It depends if we store the email address or the full path in those fields.
That being said we shouldn't have any issue even with 200 for example as no such long address exist :)

Polpetta: brb omw to buy `test@thelongestwebnameevercreatedfrfrnocapletsbreakthewholeinternetwhynotlmaooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooooo.com`

Jokes a part, keep in consideration that domain names can have emojis right now, and that these emojis are translated into ASCII characters. So I don't want to be paranoid, but if someone creates `test@😎😎😎😎⚠️⚠️⚠️⚠️🥳🥳🥳🥳🥳🔥🔥🔥🔥🔥.com` is already using a lot of characters (I don't know why but should be like 3 char per emoji? dunno)

Polpetta (commented):
I'd put 254, but then it's up to you guys! Seems fine to me other than that.

keshavbhatt:
> I'd put 254, but then it's up to you guys! Seems fine to me other than that.

Lets use 256 for now since this is what we are using everywhere, and maybe we change it everywhere once we finalize the value after internal discussion.

Polpetta (approved):
Ok let's do 256, but keep in mind that strictly speaking should be 254

frisonisland (approved):
Approve, but I'm also in favor of 254 if we store the email address and not the path as per RFC.

PR #157, opened 2023-02-20
feat: [CO-594] Add support for deeply nested attachments preview

keshavbhatt:
**What has changed:**
- parse nested part numbers
- add unit test for parsing deeply nested attachments

### DIFF
--- a/store/src/test/java/com/zimbra/cs/service/PreviewServletTest.java
+++ b/store/src/test/java/com/zimbra/cs/service/PreviewServletTest.java
@@ -1,6 +1,7 @@
 package com.zimbra.cs.service;
 
 import static com.zimbra.cs.service.PreviewServlet.IMG_THUMBNAIL_REGEX;
+import static com.zimbra.cs.service.PreviewServlet.PDF_PREVIEW_REGEX;
 import static org.junit.Assert.assertEquals;
 import static org.junit.Assert.assertTrue;
 import static org.mockito.Mockito.atLeastOnce;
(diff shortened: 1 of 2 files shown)


frisonisland (approved):
Ok to me, are we able also to mock the Preview response? Such as that for a given document we get the content displayed coorectly

keshavbhatt:
> Ok to me, are we able also to mock the Preview response? Such as that for a given document we get the content displayed coorectly

I think this should be covered in preview service codebase. Let me know why we could need it here. I mean, its perview service who decides what to send as response. This implimentation is just a wrapper to make the preview service usable for mailbox.

store/src/test/java/com/zimbra/cs/service/PreviewServletTest.java
```diff
@@ -336,6 +337,20 @@ public void shouldReturnCorrectQueryWhenCalledGenerateQuery() {
     assertEquals("png", query.getOutputFormat().get());
   }
 
+  @Test
+  public void shouldMatchNestedPartNumberFromRequestQuery() {
+    final String requestUrlForPdfPreview = "https://nbm-s01.demo.zextras.io/service/preview/pdf/14473/1.3.2.1.2.3";
```
Polpetta, line 342: Is `nbm-s01.demo.zextras.io` really necessary or can we put `example.com`?
```suggestion
    final String requestUrlForPdfPreview = "https://example.com/service/preview/pdf/14473/1.3.2.1.2.3";
```

Polpetta (changes requested):
I've only one thing I'd like to discuss/change about

Polpetta (approved):

PR #167, opened 2023-02-28
feat: [CO-573] expose soap prometheus metrics

frisonisland:
This PR adds the /metrics endpoint to expose collected SOAP metrics.

It **register a MetricsServletModule** using Guice that **serves** the Prometheus MetricServlet on **/metrics**.
**Also soap.prom dump job has been removed.**

How it works:
- **the Mailbox will expose a Summary of "soap_exec_ms"**, resulting in two counters: soap_exec_ms_count (# observations) and soap_exec_ms_sum (sum of duration of all api calls)
  - the soap_exec_ms_count equals to number of api calls 
- **the metrics will not reset when scraped, but only when the service restarts** (this allows other clients to be able to get data without interfering, also it is not time dependent), opposed to the .prom dump
- **To get the average duration (1 minute interval)** of a call you can use **sum(increase(soap_exec_ms_sum[1m])) / sum(increase(soap_exec_ms_count[1m]))**
- **To get the number of calls in an interval** you can use: **sum(increase(soap_exec_ms_count[interval]))*(interval - scrape_period)/(interval)**. 
**For 1 minute and scrape every 15 seconds: sum(increase(soap_exec_ms_count[1m])*(45/60))**
This includes a correction factor to deal with accuracy.


frisonisland:
I'm looking also for suggestion on naming, etc. The endpoint is working and exposing tracked metrics.

pom.xml
```diff
@@ -788,6 +788,137 @@
         <artifactId>freemarker</artifactId>
         <version>2.3.19</version>
       </dependency>
+      <dependency>
+        <groupId>com.google.inject.extensions</groupId>
+        <artifactId>guice-servlet</artifactId>
```
Polpetta, line 793: Are my eyes REALLY reading this? :eyes: Am I dreaming?

pom.xml
```diff
@@ -788,6 +788,137 @@
...
+        <version>1.33</version>
+      </dependency>
+      <dependency>
+        <groupId>org.javassist</groupId>
+        <artifactId>javassist</artifactId>
+        <version>3.21.0-GA</version>
+      </dependency>
+      <dependency>
+        <groupId>com.google.inject</groupId>
+        <artifactId>guice</artifactId>
```
Polpetta, line 808: image

Polpetta (approved):
92 minutes of :clap: [cit]

pom.xml
frisonisland: Lol, don't get too excited :laughing: , it's just some :tropical_drink (juice): . 
There is nothing injected yet, it works for this use case but adding the same behavior to the other existing servlets will be hard.

aheeva-yuliya (approved):

store/src/main/java/com/zextras/mailbox/servlet/MetricsServletModule.java
```diff
@@ -0,0 +1,15 @@
...
+import io.prometheus.client.exporter.MetricsServlet;
+
+public class MetricsServletModule extends ServletModule {
+
+  @Override
+  protected void configureServlets() {
+    CollectorRegistry.defaultRegistry.register(new SoapCollector());
+    serve("/metrics").with(new MetricsServlet());
+  }
+}
```
frisonisland, line 15: Btw I have found a more idiomatic way to define this, using injection
```suggestion
  @Override
  protected void configureServlets() {
    serve("/metrics").with(MetricsServlet.class);
  }

  @Provides
  @Singleton
  public SoapCollector provideSoapCollector() {
    final SoapCollector soapCollector = new SoapCollector();
    return soapCollector;
  }

  @Provides
  @Singleton
  public CollectorRegistry provideCollector() {
    return CollectorRegistry.defaultRegistry;
  }

  @Provides
  @Singleton
  public MetricsServlet provideMetricsServlet(CollectorRegistry collectorRegistry, SoapCollector soapCollector) {
  collectorRegistry.register(soapCollector);
    return new MetricsServlet(collectorRegistry);
  }
```
If nobody has nothing against current changes I will consider adding it in the next iteration :sunglasses:

pom.xml
```diff
@@ -788,6 +788,137 @@
...
+      <dependency>
+        <groupId>org.glassfish.hk2.external</groupId>
+        <artifactId>javax.inject</artifactId>
+        <version>2.5.0-b42</version>
+      </dependency>
+      <dependency>
+        <groupId>org.glassfish.hk2</groupId>
+        <artifactId>hk2-api</artifactId>
+        <version>2.5.0-b42</version>
+      </dependency>
```
keshavbhatt, line 921: All these dependencies really required or we added them in advance?

keshavbhatt (approved):

pom.xml
frisonisland: No, they are just declaration on <dependencyManagement> but still not required, let me fix that.

frisonisland:
The current CSV dump system is interferring with the exporter and clearing data, I have to remove the csv exporter logic also

store/src/main/java/com/zimbra/cs/stats/ZimbraPerf.java
```diff
@@ -574,7 +574,6 @@ private static void initializeForMainZimbraJVM() {
 
       // CSV
       statsScheduler.schedule(new CsvStatsDumper(mailboxdTracker), DUMP_FREQUENCY);
-      statsScheduler.schedule(new CsvStatsDumper(SOAP_TRACKER), DUMP_FREQUENCY);
```
keshavbhatt, line 577: Is this intentional? If I understood correctly, we want to keep the csv stats?

keshavbhatt (commented):
Dropped a comment about csv based stat collection.
Also will these metrics accessible directly under /metrics endpoint? What about other classes of metrics?

store/src/main/java/com/zimbra/cs/stats/ZimbraPerf.java
frisonisland: Thanks, I forgot about this, re-adding csv dump.

PR #169, opened 2023-03-02
feat: [CO-619] store Let's encrypt certs to LDAP before proxy conf generation 

aheeva-yuliya:
What's changed: 
I've removed "expand" flag from the request (was needed to understand if domain admin asking to issue a new LE cert bcz a virtual host name was added or deleted from the domain) - instead we will always use **--cert-name** as it gives more control over the certificate (will be automaticlly checked by certbot if the cert is needed to be modified somehow. If all argumenets mathches it will keep existing cert if it's still valid othervise would modify existing cert as needed (add or remove VHN or modify public service host name if it was changed). Certbot will store a cert under working directory with the domain name passed with **--cert-name** so it could be easily found and understood to which domain cert belongs to.

So the flow is:
1) admin requests a certficate => 
2) certbot isuues a cert and stores it under working directory in the folder with **domain name** => 
3) in order to activate LE certs who(what)ever needs to execute proxyconfgen and reload process
proxyconfgen process gets all domains in the infrastructure, for each checks (using a domain name) if a cert exists for this domain inside the certbot working dir, if yes will modify the values for the domain cert and priv key in ldap and continue its job for generating proxy configuration

So a domain admin will see LE cert enabled and could retrieve it from ldap after proxyconfgen will be execuated globally (which he would be notified about) - is not covered by this pr

### DIFF
--- a/store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
+++ b/store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
@@ -6,12 +6,15 @@
 package com.zimbra.cs.util.proxyconfgen;
 
 import com.zimbra.common.account.Key;
+import com.zimbra.common.account.Key.DomainBy;
 import com.zimbra.common.account.ZAttrProvisioning;
 import com.zimbra.common.service.ServiceException;
 import com.zimbra.common.util.CliUtil;
 import com.zimbra.common.util.Log;
 import com.zimbra.common.util.LogFactory;
 import com.zimbra.common.util.StringUtil;
+import com.zimbra.cs.account.AccountServiceException;
+import com.zimbra.cs.account.Domain;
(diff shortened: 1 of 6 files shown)


store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -564,7 +510,7 @@ private static void addCustomTemplateWarningHeader(
    * @author Jiankuan
    * @deprecated use expandTemplateByExplodeDomain instead
    */
-  @SuppressWarnings("DeprecatedIsStillUsed")
+  //DeprecatedIsStillUsed
```
Polpetta, line 513: If you don't think this comment is necessary anymore remove it ;)

Polpetta (approved):
LGTM

frisonisland (approved):

PR #170, opened 2023-03-06
fix: [CO-553] Calendar invites parsing in Outlook

keshavbhatt:
**What has changed:**
- fix mime-parsing logic by correctly order mime parts in multi-mime part component of invite email
    - mime part with content-type text/plain are placed at top in multi-mime parts with content-type mimepart/alternatives in calendar invite emails
- general code cleanup

### DIFF
--- a/store/src/main/java/com/zimbra/cs/service/mail/ParseMimeMessage.java
+++ b/store/src/main/java/com/zimbra/cs/service/mail/ParseMimeMessage.java
@@ -5,35 +5,9 @@
 
 package com.zimbra.cs.service.mail;
 
-import java.io.ByteArrayInputStream;
-import java.io.ByteArrayOutputStream;
-import java.io.IOException;
-import java.io.InputStream;
-import java.io.OutputStream;
-import java.io.UnsupportedEncodingException;
-import java.util.ArrayList;
-import java.util.Collections;
-import java.util.HashMap;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-
-import javax.activation.DataHandler;
-import javax.mail.MessagingException;
-import javax.mail.Part;
-import javax.mail.SendFailedException;
-import javax.mail.internet.InternetAddress;
-import javax.mail.internet.InternetHeaders;
-import javax.mail.internet.MimeBodyPart;
-import javax.mail.internet.MimeMessage;
-import javax.mail.internet.MimeMultipart;
-import javax.mail.internet.MimePart;
-import javax.mail.internet.MimeUtility;
-import javax.mail.util.SharedByteArrayInputStream;
-
-import com.google.common.base.Charsets;
 import com.google.common.base.Strings;
 import com.google.common.collect.ImmutableSet;
+import com.zimbra.common.account.ZAttrProvisioning;
 import com.zimbra.common.calendar.ZCalendar;
 import com.zimbra.common.localconfig.LC;
 import com.zimbra.common.mailbox.ItemIdentifier;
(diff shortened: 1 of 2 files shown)


store/src/main/java/com/zimbra/cs/service/mail/ParseMimeMessage.java
```diff
@@ -84,936 +58,1113 @@
...
+        if (subpart == plainTextPart.orElse(null) || subpart == htmlTextPart.orElse(null)) {
+          continue;
+        }
+        setContent(mm, mmpNew, subpart, null, ctxt, isQPencodeRequired);
+      }
+
+      // add htmlText part as last element
+      if (htmlTextPart.isPresent()) {
+        setContent(mm, mmpNew, htmlTextPart.get(), null, ctxt, isQPencodeRequired);
+      }
```
frisonisland, line 632: **I'm trying to understand the flow and logic**: I read about multipart/mixed and alternative here https://www.w3.org/Protocols/rfc1341/7_2_Multipart.html  
**Do we want to add plaintext and html mimepart content, as alternative ways to represent the Calendar, as multipart/alternative children to make the client able to choose how it wants to display calendar information?**

Unfortunately the methods to parse the content are static and undocumented, I would prefer having  elem.setContent() rather than setContent(...).
Some things that could help:
- **Consider adding a brief description of the incoming call and parsing flow, desired output.** (Example: the message will be sent with multipart/alternative element at top, then the text/html content alongside text/plain, etc.)
- **Consider documenting the static methods** about setting content, as **it is not clear what they perform without having to read all the code**
- **If not too much trouble consider refactoring to avoid static methods and prefer the instance.setContent() approach.** 

Thanks :pray:

keshavbhatt: Hi Davide, Thanks for your feeedback. :pray: 

> * **Consider adding a brief description of the incoming call and parsing flow, desired output.** (Example: the message will be sent with multipart/alternative element at top, then the text/html content alongside text/plain, etc.)

You can find javadoc on the method that i introduces to fix the issue, with reason and explaination what I did to fix the bug in PR's description already. I hope you understand that scope of this issue is to fix a bug, and not to understand everything & document it. (I am afraid to put wrong information as the code is mess and hard to understand). I hope we will spend some time in future to re-write this whole logic investigating and learning a lot more about the topic.

just to quickly answer your query since you asked:  
- Sending a Calendar invite just requires a well formatted iCalendar object within your invite email. 
- To describe the email and add additional information about the event we add information in different forms to make cleint parse it and show it to users.
- This information is organized and added in form of Mimeparts which are attached to MimeMessage, the attached mime parts can hold different type of content which is described by setting Content-type attribute in them.
- Out Calendar invite email consist of MimeMessage which includes mimparts inisde it witht he following hierarchy: 
```
MimeMessage
   multipart/alternative
     text/plain
     text/calendar
     text/html
```
Multipart/alternative is used to indicate that there are multiple representations of the same content. We send three representations in Carbnonio as mentioned above (some calender servers also sends base64 representation of this data too). Client who parse the email then choose which representation to display based on their capabilities and the preferences.

> * **Consider documenting the static methods** about setting content, as **it is not clear what they perform without having to read all the code**

Not covered in the scope of task. 

> * **If not too much trouble consider refactoring to avoid static methods and prefer the instance.setContent() approach.**

I have only added one static method in this Utility Class. The method in question has lots of logic and has nothing to do with the scope of this task.

frisonisland: > * Sending a Calendar invite just requires a well formatted iCalendar object within your invite email.
> * To describe the email and add additional information about the event we add information in different forms to make cleint parse it and show it to users.
> * This information is organized and added in form of Mimeparts which are attached to MimeMessage, the attached mime parts can hold different type of content which is described by setting Content-type attribute in them.
> * Out Calendar invite email consist of MimeMessage which includes mimparts inisde it witht he following hierarchy:
> 
> ```
> MimeMessage
>    multipart/alternative
>      text/plain
>      text/calendar
>      text/html
> ```
> 
> Multipart/alternative is used to indicate that there are multiple representations of the same content. We send three representations in Carbnonio as mentioned above (some calender servers also sends base64 representation of this data too). Client who parse the email then choose which representation to display based on their capabilities and the preferences.
> 
Thanks! Your comment explains and confirms the fix you are bringing with this PR. :slightly_smiling_face: 
I agree to do other major changes in another PR, what matters here to me is the reason and logic of the fix!

frisonisland (approved):

Polpetta (approved):

PR #172, opened 2023-03-09
feat: [CO-611] prom mailbox JVM stats

frisonisland:
This PR adds JVM prom metrics:
- **Hotspot metrics**: https://prometheus.github.io/client_java/io/prometheus/client/hotspot/package-summary.html
- Remove MemoryStats class and old "legacy" jvm stats

I have also added Copyright to files


store/src/main/java/com/zextras/mailbox/servlet/MetricsServletModule.java
```diff
@@ -1,15 +1,66 @@
...
   protected void configureServlets() {
-    CollectorRegistry.defaultRegistry.register(new SoapCollector());
-    serve("/metrics").with(new MetricsServlet());
+    serve("/metrics").with(MetricsServlet.class);
+  }
+
+  /**
+   * Provides prometheus {@link MetricsServlet} with the defined collector registry.
+   *
+   * @param collectorRegistry thge registry holding collectors
```
keshavbhatt, line 29:
```suggestion
   * @param collectorRegistry the registry holding collectors
```

keshavbhatt (changes requested):
Changes looks good to me, 

**Only concerns:**
- What about the JMX metrics collection that is already in place? Having two different mechanism to do same task will imapct the performance for sure. I will remove the existing JXM one and all related logic to avoid the overhead. 
- What about removing the exisitng .prom stat collection? I think we should get rid of that too.
- Also if we want to keep the .csv based jxm scrapped stats, consider migrating it to use stats scraped from this one.

frisonisland:
> Changes looks good to me,
> 
> **Only concerns:**
> 
> * What about the JMX metrics collection that is already in place? Having two different mechanism to do same task will imapct the performance for sure. I will remove the existing JXM one and all related logic to avoid the overhead.
> * What about removing the exisitng .prom stat collection? I think we should get rid of that too.
> * Also if we want to keep the .csv based jvm stats, consider migrating it to use stats scraped from hotspot.

I would propose removing the existing mechanism, although it would be hard to integrate the new system with the CSV, because the old legacy code does not use injection. I don't see the point in having the csv file.
Also the new code is a sort of a blackbox: we use prometheus client libraries to do the magic, for getting source data to outputting, there is no parsing logic.

What do you think?

keshavbhatt:
> I don't see the point in having the csv file

Same here, just remove them.

aheeva-yuliya (approved):

Polpetta (approved):

Polpetta:
I'd keep everything as KISS as possible

keshavbhatt (approved):

PR #175, opened 2023-03-13
feat: [CO 621] Allow LE certificates to be generated from Mailbox endpoint

aheeva-yuliya:
What's changed:

- add async to remote execution
- start notifying global/domain recipients
- change certbot working dir bcz of updated certbot package 
- slightly modify some old tests ("It ain't much, but it's honest work.") it won't help but better to run init server and clean data with every test

Current flow:
1) the admin requests to issue cert with api 
2) mailbox starts checking everything what’s possible to check (if smth is wrong will thow an exeption)
3) if everything is ok will send a response 
<message domain="test.zextras.io">The System is processing your certificate generation request. It will send the result to the Domain notification recipients.</message> 
4) meanwhile remote manager executing certbot command and recieve a message from it

we have 4 scenarios : №1 system failures (not related to certbot), №2 certbot failures, №3 certbot success (certificate recieved or certificate not yet due for renewal),  №4 other (smth is wrong and wasn’t expected)

5) we notifying global about first 3 with the messages we got (not modifying them)
6) we notifying domain about №2 and about №3 (only if certificate was recieved) with the templates provided by Danielle Veri
7) №4 stored under opt/zextras/log/mailbox.log

Related PRs:
 https://github.com/Zextras/carbonio-core-utils/pull/38
 https://github.com/Zextras/carbonio-build/pull/50
 https://bitbucket.org/zextras/carbonio-core/pull-requests/42


store/src/main/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -246,6 +249,27 @@ public MailSender setSession(Account account) throws ServiceException {
...
 
+    /**
+     * Sets an alternate JavaMail <tt>Session</tt> and SMTP hosts
+     * that will be used to send the message based on the domain.
+     * The default behavior is to use SMTP settings from
+     * the <tt>Session<tt> on the {@link MimeMessage}.
+     * @throws ServiceException if not able to get SMTP session for the domain
+     *
+     * @author Yuliya Aheeva
+     * @since 23.4.0
```
frisonisland, line 260: Remember to put 23.5.0 ;)

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,326 @@
...
+  }
+
+  /**
+   * Creates a map based on domain values and Remote Manager/Certbot output which would be used to
+   * create {@link javax.mail.internet.MimeMessage}.
+   *
+   * @param outputMessage output from RemoteManager/Certbot
+   * @return map with needed values of FROM, TO, SUBJECT and MESSAGE TEXT
+   */
+  protected static Map<String, Object> createIssueCertNotificationMap(
```
frisonisland, line 143: What is the goal of this method? To add from: and to: in form of map? Asking because it seems a bit generic, also all values are String but we return a Map of <String,Object> and later cast again to String.
I suggest to return String, not Object if we deal with only Strings. 
Edit: I see it deals with lists. Maybe we can create a method for each thing you need (I don't like casting :slightly_smiling_face: ).
Something like: 
- List<String> getRecipients(Domain domain)
- String getNotificationSender(Domain domain), etc. 
We could return Optionals also.

store/src/main/java/com/zimbra/cs/mailbox/MailSender.java
```diff
@@ -246,6 +249,29 @@ public MailSender setSession(Account account) throws ServiceException {
...
+    public MailSender setSession(Domain domain) throws ServiceException {
+        try {
+            mSession = JMSession.getSmtpSession(domain);
+        } catch (MessagingException e) {
+            throw ServiceException.FAILURE("Unable to get SMTP session for " + domain, e);
+        }
+        mSmtpHosts.clear();
+        mSmtpHosts.addAll(JMSession.getSmtpHosts(domain));
+        return this;
+    }
```
frisonisland, line 273: Is this safe? What I mean is I don't like the ability to change the session after some operation are already done. 
Also I don't see the need of creating a MailSender without session, consider creating a factory method and a private constructor with session and hosts maybe. What do you think?

frisonisland (changes requested):

store/src/main/java/com/zimbra/cs/mailbox/MailSender.java
aheeva-yuliya: if you check MailSender it has  public MailSender setSession(Account account) which is the same as this one
im getting requested mailsender from the mailbox by providing the domain 
it is the same logic as SendMsg handler doing but there is getting it based on account

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
aheeva-yuliya: the goal of this method is to undersnad which message to which recipients should be sent
bcz we need to send some messages for domain recipients and some only to global
it deals with string and string[] 
i will think mb of little refactor to avoid casting ok

store/src/main/java/com/zimbra/cs/mailbox/MailSender.java
frisonisland: Ok! Then I have an issue the original approach :sweat_smile: 
I think the session should be final, consider adding a "TODO: make session final to avoid issues" if you think it could lead to problems.

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
aheeva-yuliya: @frisonisland i've refactored it No more casting  slightly_smiling_face

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,326 @@
...
+              SUCCESS_DOMAIN_NOTIFICATION_TEMPLATE.replace("<DOMAIN_NAME>", domain.getName()),
+              expire);
+
+      notificationMap.put(DOMAIN_MESSAGE, domainMessage);
+      notificationMap.put(DOMAIN_FROM, domainFrom);
+      notificationMap.put(DOMAIN_TO, domainTo);
+    }
+
+    // in any other cases (like certificate is not yet due for renewal ... etc)
+    // only global admin would be notified
```
keshavbhatt, line 199: How this comment is relevant here? :eyes:

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,312 @@
...
+    notificationMap.put(SUBJECT_RESULT, CERTBOT_SUCCESS);
+    return notificationMap;
+  }
+
+  /**
+   * Creates {@link javax.mail.internet.MimeMessage} list what would be sent to recipients.
```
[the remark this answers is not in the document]
frisonisland: The method, if I got it, creates  a global message and a domain message, if present. We should state it in the description. It's not clear if we always set global and domain message or we can end up having only the domain message because of the catch() logic. Th subject is the same for both emails, right?

frisonisland (changes requested):
I think the whole map logic is a bit confusing. Maybe consider a more streamlined version where you just do, from the notify:
- parse output
- create  global mime message
- create domain mime message
- send both messages using the same sender
This way we avoid having if conditions in the same method to handle domain and global case.
it is also easier to spot the wanted behavior.

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,326 @@
...
+          + "The certificate for <DOMAIN_NAME> was successfully received.\n"
+          + "Please NOTE  that the Certificate and Key will be available after the proxy reload.\n"
+          + "You’ll be able to download them from the Certificate section in the admin interface.\n"
+          + "\n"
+          + "The files will be automatically updated when the certificate renews.\n";
+
+  private static final Provisioning provisioning = Provisioning.getInstance();
+
+  private CertificateNotificationManager() {
+    throw new RuntimeException("CertificateNotificationManager class cannot be instantiated.");
```
keshavbhatt, line 81:
```suggestion
    throw new UnsupportedOperationException("CertificateNotificationManager class cannot be instantiated.");
```

store/src/test/java/com/zimbra/cs/service/admin/IssueCertTest.java
```diff
@@ -108,37 +109,55 @@ public void clearData() {
...
+        mail,
+        AdminConstants.DEFAULT_CHAIN,
+        domainName,
+        publicServiceHostName,
+        expectedDomain.getVirtualHostname())).thenReturn(expectedCommand);
 
     final Element response = handler.handle(request, context);
     final Element message = response.getElement(E_MESSAGE);
+
     assertEquals(message.getAttribute(A_DOMAIN), domainName);
```
keshavbhatt, line 157:
```suggestion
    assertEquals(domainName, message.getAttribute(A_DOMAIN));
```
correctly order expected and actual values.

store/src/test/java/com/zimbra/cs/service/admin/IssueCertTest.java
```diff
@@ -108,37 +109,55 @@ public void clearData() {
...
-    assertEquals(message.getText(), expectedMessage);
+    assertEquals(message.getText(), IssueCert.RESPONSE);
```
keshavbhatt, line 158:
```suggestion
    assertEquals(IssueCert.RESPONSE, message.getText());
```
correctly order expected and actual values.

keshavbhatt (changes requested):
I added some suggestions and requested some changes.

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,312 @@
...
+  }
+
+  /**
+   * Creates a map based on Remote Manager/Certbot output which would be used to
+   * create {@link javax.mail.internet.MimeMessage}.
+   *
+   * @param outputMessage output from RemoteManager/Certbot
+   * @return map with needed values of notification SUBJECT and MESSAGE TEXT
+   */
+  protected static Map<String, String> createIssueCertNotificationMap(
```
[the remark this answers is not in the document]
keshavbhatt: Also, `protected` makes no sense here, as this utility class is not ment to be subclassed (to enforce this add `final` in class definition). Then we can remove protected and make it package private by  just defining it like:
`static Map<String, String> createIssueCertNotificationMap(){}` ...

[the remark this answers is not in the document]
frisonisland: IMHO not even package private, this is not meant to be shared. Ok protected in case we need to change its behavior (but I assume not/does not make sense)

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,326 @@
...
+          + "wasn't able to verify the validity of the domain.\n"
+          + "\n"
+          + "Most common reasons that could cause this kind of failure are:\n"
+          + "- Misspelled or missing Public Service Hostname and/or Virtual Hostname."
+          + " Make sure both are filled in with a valid Fully Qualified Domain Name.\n"
+          + "- Wrong or missing A/AAAA entry for Public Service Hostname and/or Virtual Hostname."
+          + " Make sure there is a valid, public resolution for the Fully Qualified Domain Name"
+          + " used for any of the Public Service Hostname or Virtual Hostname.\n"
+          + "- Private or unreachable IP address. In order to validate the domain name,"
+          + " the certificator must be able to resolve and browse the FQDN provided.\n"
```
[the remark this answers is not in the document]
aheeva-yuliya: it was provided by luca and danielle, but i will change

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,312 @@
...
+   *
+   * @param session {@link javax.mail.Session}
+   * @param notificationMap map with needed values MESSAGE TEXT
+   * @return a list of {@link javax.mail.internet.MimeMessage}
+   * @throws ServiceException if unable to parse addresses or create a MimeMessage
+   */
+  protected static List<MimeMessage> createMimeMessageList(
+      MailSender sender, Domain domain, Map<String, String> notificationMap) throws ServiceException {
+
+    Session session = sender.getCurrentSession();
```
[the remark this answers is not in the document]
aheeva-yuliya: ok will get an object of CertificateNotificationManager

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,312 @@
...
+
+  /**
+   * Creates {@link javax.mail.internet.MimeMessage} list what would be sent to recipients.
+   *
+   * @param session {@link javax.mail.Session}
+   * @param notificationMap map with needed values MESSAGE TEXT
+   * @return a list of {@link javax.mail.internet.MimeMessage}
+   * @throws ServiceException if unable to parse addresses or create a MimeMessage
+   */
+  protected static List<MimeMessage> createMimeMessageList(
```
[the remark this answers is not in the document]
aheeva-yuliya: ok getting an object

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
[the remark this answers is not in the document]
aheeva-yuliya: changed it

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
aheeva-yuliya: outdated

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,278 @@
...
+      Mailbox mbox, Domain domain) {
+    return new CertificateNotificationManager(mbox, domain);
+  }
+
+  /**
+   * Notifies global and domain recipients about certificate generation result.
+   *
+   * @param outputMessage a message returned by RemoteManager/Certbot acme client
+   */
+  public void notify(String outputMessage) {
```
frisonisland, line 109: Just one thing about this method: I think DOMAIN_MESSAGE and GLOBAL_MESSAGE do not add anything to the logic, instead they are not necessary and lead to double conditions, e.g.: 
- _if failure -> add DOMAIN_MESSAGE, if DOMAIN_MESSAGE add domain mime message_ 
VS 
- _if failure -> add mime message_.

I suggest also:
```java
/**
   * Notifies global and domain recipients about certificate generation result.
   * In case of failure only global admin receives the message.
   * In case of certbot failure both global and domain admin should receive the status message
   * In case of success global admin receives the message and domain admin receives it only if certbot request was completed.
   *
   * @param outputMessage a message returned by RemoteManager/Certbot acme client
   */
  public void notify(String outputMessage) {
    String domainName = domain.getName();

    ZimbraLog.rmgmt.info(
        "Issuing LetsEncrypt cert command for domain "
            + domainName
            + " was finished with the following result: "
            + outputMessage);


    try {
      Config config = Provisioning.getInstance().getConfig();
      String globalFrom =
          Optional.ofNullable(config.getCarbonioNotificationFrom())
              .orElseThrow(
                  () ->
                      ServiceException.FAILURE(
                          "Global CarbonioNotificationFrom attribute is not present.", null));
      String[] globalTo =
          Optional.ofNullable(config.getCarbonioNotificationRecipients())
              .orElseThrow(
                  () ->
                      ServiceException.FAILURE(
                          "Global CarbonioNotificationRecipients attribute is not present.", null));
      String domainFrom =
          Optional.ofNullable(domain.getCarbonioNotificationFrom()).orElse(globalFrom);
      String[] domainTo =
          Optional.ofNullable(domain.getCarbonioNotificationRecipients()).orElse(globalTo);

      MailSender sender = mbox.getMailSender(domain);
      List<MimeMessage> mimeMessageList = new ArrayList<>();

      if (outputMessage.contains(SYSTEM_FAILURE)) {
        mimeMessageList.add(createMimeMessage(sender.getCurrentSession(), domain.getName() + SUBJECT_TEMPLATE + SYSTEM_FAILURE,
            convert(globalFrom), convert(globalTo), outputMessage));
      }
      else if (outputMessage.contains(FAIL)) {
        mimeMessageList.add(createMimeMessage(sender.getCurrentSession(), domain.getName() + SUBJECT_TEMPLATE + CERTBOT_FAILURE,
            convert(globalFrom), convert(globalTo), outputMessage));

        mimeMessageList.add(createMimeMessage(sender.getCurrentSession(), domain.getName() + SUBJECT_TEMPLATE + CERTBOT_FAILURE,
            convert(domainFrom), convert(domainTo), createDomainFailureMessage(domain, outputMessage)));
      } else {
        // all ok
        mimeMessageList.add(createMimeMessage(sender.getCurrentSession(), domain.getName() + SUBJECT_TEMPLATE + CERTBOT_SUCCESS,
            convert(globalFrom), convert(globalTo), outputMessage));
        if (outputMessage.contains(RECEIVED)) {
          mimeMessageList.add(createMimeMessage(sender.getCurrentSession(), domain.getName() + SUBJECT_TEMPLATE + SUCCESS_RESULT,
              convert(domainFrom), convert(domainTo), createDomainSuccessMessage(domain, outputMessage)));
        }
      }
      // parse the message, add keys based on message
      sender.sendMimeMessageList(mbox, mimeMessageList);

      ZimbraLog.rmgmt.info(
          "Notifications about LetsEncrypt certificate generation were sent "
              + "for the global and domain "
              + domainName
              + " recipients.");

    } catch (Exception e) {
      ZimbraLog.rmgmt.info(
          "Notifications about LetsEncrypt certificate generation weren't sent "
              + "for the global and domain "
              + domainName
              + " recipients.\n"
              + "Sending failure: "
              + e.getMessage());
    }
  }

  private String createDomainFailureMessage(Domain domain, String outputMessage) {
    return String.join(
        "",
        HEADER,
        FAILURE_RESULT,
        FAILURE_DOMAIN_NOTIFICATION_TEMPLATE.replace("<DOMAIN_NAME>", domain.getName()));
  }
  String expire = "\n";

    String regex = "\\d{4}-\\d{2}-\\d{2}";
    Matcher matcher = Pattern.compile(regex).matcher(outputMessage);
    if (matcher.find()) {
      String expiresTemplate = "This certificate expires on ";
      expire = String.join("",expire, expiresTemplate, matcher.group(), ".");
    }
    return
        String.join(
            "",
            HEADER,
            SUCCESS_RESULT,
            SUCCESS_DOMAIN_NOTIFICATION_TEMPLATE.replace("<DOMAIN_NAME>", domain.getName()),
            expire);

  }
``` 
Again I just think the **parseOutput** is a bit generic in naming + you don't need a map. That approach reminds me of Python :sweat_smile:  
Also explain branches conditions in JavaDoc as it defines the behavior.

aheeva-yuliya: @frisonisland i've already modify it
check my last commit pls

keshavbhatt (approved):

store/src/main/java/com/zimbra/cs/rmgmt/RemoteCertbot.java
```diff
@@ -69,11 +86,28 @@ public String createCommand(String remoteCommand, String email, String chain,
...
+   * @param notificationManager an object of {@link com.zimbra.cs.service.admin.CertificateNotificationManager}
+   * @param command a Certbot command to be executed remotely
+   *
+   * @author Yuliya Aheeva
+   * @since 23.5.0
+   */
+  public void supplyAsync(CertificateNotificationManager notificationManager, String command) {
+    CompletableFuture.supplyAsync(() -> execute(command))
+        .thenApply(notificationManager::createIssueCertNotificationMap)
+        .thenAccept(notificationManager::notify);
```
Polpetta, line 102: Is there any global thread pool where to exec this task instead of putting it in the default ForkJoinPool? Kinda makes it hard to test

store/src/main/java/com/zimbra/cs/service/admin/CertificateNotificationManager.java
```diff
@@ -0,0 +1,290 @@
...
+      sender.sendMimeMessageList(mbox, mimeMessageList);
+
+      ZimbraLog.rmgmt.info(
+          "Notifications about LetsEncrypt certificate generation were sent "
+              + "for the global and domain "
+              + domainName
+              + " recipients.");
+
+    } catch (Exception e) {
+      ZimbraLog.rmgmt.info(
```
Polpetta, line 174: Could be better to issue a warning than an info message?

store/src/test/java/com/zimbra/cs/mailbox/ContactTest.java
```diff
@@ -415,20 +420,15 @@ public void testTruncatedContactsTgzImport() throws IOException {
       try {
         ArchiveFormatter.readArchiveEntry(ais, aie);
       } catch (IOException e) {
-        e.printStackTrace();
+        System.out.println("Expected exception was caught: " + e);
```
Polpetta, line 423: In this case, please use the self4j logger that can be customized outside the tests in order to be more or less verbose

`private final Logger log = LoggerFactory.getLogger(this.getClass())` at the beginning of the test run and you're good to log tests how much you want

Polpetta (changes requested):
Some minor details to fix, the implementation seems ok!

Polpetta (approved):

PR #178, opened 2023-03-14
feat: [CO-643] Make Carbonio NotificationRecipients/From domainInherited

keshavbhatt:
**What has changed:**
- make carbonioNotificationRecipients and carbonioNotificationFrom inherit from global config

We missed this in the [CO-539], as it was not asked in the requirements, but it makes sense to have the domain entries inherit value from global config if they don't have these two attributes set.

This change requires no migration as 23.4.0 is not yet released.

### DIFF
--- a/store/conf/attrs/attrs.xml
+++ b/store/conf/attrs/attrs.xml
@@ -9992,11 +9992,11 @@ TODO: delete them permanently from here
   <globalConfigValue>https://www.zextras.com</globalConfigValue>
 </attr>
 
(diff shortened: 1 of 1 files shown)


frisonisland (approved):
LGTM, If I understand "domainInherited" will make the value inherited from global config when not set on the domain

keshavbhatt:
> LGTM, If I understand "domainInherited" will make the value inherited from global config when not set on the domain

Yes.

aheeva-yuliya (approved):

PR #179, opened 2023-03-15
chore: [CO-644] mailbox service-discover registration

frisonisland:
Uniforms mailbox service registration to: https://github.com/Zextras/consul-templates/tree/main/services-package

Tests:
 - installation ok
 - setup ok
 - token is now under /etc/carbonio/mailbox/service-discover/token
 **- copy old token to new in postinst. This in case other services are restarted and need the new token but pending-setups did not run (although that shouldn't happen, the token should be read only by the service itself, I'm a bit confused about this)**
 - old token /etc/zextras/carbonio-mailbox/token still present in the system (any idea how we can remove it? automatically or manually?)

### DIFF
--- a/packages/appserver-service/PKGBUILD
+++ b/packages/appserver-service/PKGBUILD
@@ -54,7 +54,7 @@ postinst() {
 
   usermod -a -G 'carbonio-mailbox' 'zextras'
 
-  # Copy token old token. This way services using new token can be restarted and use the new one
+  # Copy old token. This way services using new token can be restarted and use the new one
   # without the need of running pending-setups before
   if [ -f /etc/zextras/carbonio-mailbox/token ]; then
     cp /etc/zextras/carbonio-mailbox/token /etc/carbonio/mailbox/service-discover/token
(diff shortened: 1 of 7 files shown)


packages/appserver-service/PKGBUILD
```diff
@@ -46,8 +54,11 @@ postinst() {
 
   usermod -a -G 'carbonio-mailbox' 'zextras'
 
-  mkdir -p '/etc/zextras/carbonio-mailbox/'
-  chown 'carbonio-mailbox:carbonio-mailbox' '/etc/zextras/carbonio-mailbox/'
+  # Copy token old token. This way services using new token can be restarted and use the new one
```
keshavbhatt, line 57:
```suggestion
  # Copy old token. This way services using new token can be restarted and use the new one
```

keshavbhatt (approved):
added a suggestion. 
everything else LGTM.

Polpetta (approved):

aheeva-yuliya (approved):

PR #181, opened 2023-03-16
fix: remove old token copy logic

frisonisland:
Reasons:
- Copying token could result in overwriting new one
- After package installation old token is still there for running processes
- Updated services will use new token only after restart
- Service restart should be done after pending-setups in all cases

### DIFF
--- a/packages/appserver-service/PKGBUILD
+++ b/packages/appserver-service/PKGBUILD
@@ -54,10 +54,9 @@ postinst() {
 
   usermod -a -G 'carbonio-mailbox' 'zextras'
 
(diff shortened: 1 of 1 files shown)


keshavbhatt (approved):
lgtm, just one req: please make title of merge commit more informative, to reflect what token we talking about :handshake:

Polpetta (approved):
Don't forget to add the issue number, if it is present of course!

aheeva-yuliya (approved):

frisonisland:
> Don't forget to add the issue number, if it is present of course!

I have decided to keep it within the same ticket since changes are in same devel cycle.

PR #192, opened 2023-03-31
chore: remove tag from CI

frisonisland:
**TAG options from CI does not work properly**, because on checkout it does not get all commit history, thus making changelog not working. **I propose to remove it to avoid accidents.**
Since we do the release "manually" another tool, what is left is "release-it" config to automate:
- version bump
- CHANGELOG
- push

The flow of release will be in future, roughly: 
- merge devel to main
- create release branch + execute release-it
- merge release branch into main
- TAG
- Build tag
- Merge main to devel

### DIFF
--- a/.release-it.json
+++ b/.release-it.json
@@ -1,9 +1,9 @@
 {
   "git": {
-    "requireUpstream": true,
+    "requireUpstream": false,
     "commitMessage": "chore: release ${version}",
     "tagName": "${version}",
-    "tag": true,
+    "tag": false,
     "push": true
   },
   "plugins": {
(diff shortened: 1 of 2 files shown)


keshavbhatt (approved):

Polpetta (approved):

Polpetta:
Keep in mind that you can always retrieve the whole git history by issuing `git fetch --unshallow` during CI time. Alternatively, you can configure the clone accordingly with your preferences, see this https://www.jenkins.io/doc/book/pipeline/syntax/#options `skipDefaultCheckout` to skipping the default checkout and defining your own, and https://www.jenkins.io/doc/pipeline/steps/workflow-scm-step/ for defining how to clone. If you type `git` in here https://www.jenkins.io/doc/pipeline/steps/ you get other functionalities (also commit listing/changelog, etc)

PR #200, opened 2023-04-14
chore: [CO-603] remove deprecated attributes and features

frisonisland:
# Goal
Main goal is removing deprecated attributes.
Since many attributes are still used in some features, also features that are not needed anymore have been removed. 

# Removed features
Aside attributes, these are removed features:
- Wiki
- MigrateAccount (API)
- LdapUpgrade mechanism (bugs)
- Child Account (old feature)
- isParentOf logic for child account (ACL)
- remove RunUnitTests API
- remove **most of** QA unittests classes (except 2, MessageBuilder and TestUtil)

# Other deprecated attributes
These are additional non-requested attributes that I removed because deprecated:
- "735" name="zimbraReverseProxyInactivityTimeout"
- "847" name="zimbraMobilePolicyDeviceEncryptionEnabled"
- "1070" name="zimbraDomainMandatoryMailSignatureText"
- "1071" name="zimbraDomainMandatoryMailSignatureHTML"
- "1208" name="zimbraMailEmptyFolderBatchThreshold"
- "1378" name="zimbraReverseProxyUpstreamServers"
- "1538" name="zimbraMtaStpdSoftErrorLimit"

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -69,6 +69,11 @@ pipeline {
                     env.CONTAINER_ID = sh(returnStdout: true,
                     script: "docker run -dt ${NETWORK_OPTS} carbonio-builder").trim()
                 }
+                withCredentials([usernamePassword(credentialsId: 'artifactory-jenkins-gradle-properties-splitted',
+                                                  usernameVariable: 'ARTIFACTORY_USER', passwordVariable: 'ARTIFACTORY_PWD')]) {
+                sh 'echo "artifactory_user=${ARTIFACTORY_USER}" >> build.properties'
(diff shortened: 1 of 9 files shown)


frisonisland:
While working on this I found there are some migration logic on Mailbox.java class. I removed the Wiki migration from it. I think it is safe to remove also all other migrations. Looking for feedback!

keshavbhatt:
> While working on this I found there are some migration logic on Mailbox.java class. I removed the Wiki migration from it. I think it is safe to remove also all other migrations. Looking for feedback!

I think its fine to remove everything that we dont use. The old migration mechanisms will cause no harm if it not used anymore.

PR #202, opened 2023-04-14
chore: remove Mailbox auto-update logic

frisonisland:
This is a thing I think we do not need anymore


Polpetta (approved):

keshavbhatt:
This should be fine if these migration are now part of core schema(which makes thier existence stale here). If not this could cause issues. We should confirm this and act accordingly. Also will it be possible to get rid of MailboxUpgrade  and related classes as well?

PR #203, opened 2023-04-18
feat(CO-629): remove core briefcase functionalities

Polpetta:


frisonisland:
I have found we can also probably remove this: https://github.com/Zextras/carbonio-mailbox/tree/devel/store/data/briefcase/user1

Polpetta:
We could, but I don't really now since I didn't find anything related to it with IntelliJ search, so I don't know if removing it would remove some resources used on other tests too

frisonisland (approved):

frisonisland:
> We could, but I don't really now since I didn't find anything related to it with IntelliJ search, so I don't know if removing it would remove some resources used on other tests too

Ok, no worries, I will check if it can be removed on another PR

aheeva-yuliya (approved):

PR #210, opened 2023-04-28
fix: [CO-695] enable async support in guice servlet filter

keshavbhatt:
**What has changed:**

- this enables asynchronous processing of HTTP requests. 

In this particular case NoOpRequests were failing due to missing async support on this filter. 

For more details refer: CO-695 on jira.


Polpetta:
Aren't we taking more commits than the fix itself?

keshavbhatt:
> Aren't we taking more commits than the fix itself?

We want it as hotfix now: Please consider this PR.
https://github.com/Zextras/carbonio-mailbox/pull/211

PR #220, opened 2023-05-10
feat: [CO-675] Let admin set arbitrary PublicService & Virtual hostname

aheeva-yuliya:
Now global admin and delegated admin with granted right `modifyGlobalConfig` could set an arbitrary value for zimbraPublicServiceHostname and zimbraVirtualHostname.

### DIFF
--- a/store/src/main/java/com/zimbra/cs/service/admin/ModifyDomain.java
+++ b/store/src/main/java/com/zimbra/cs/service/admin/ModifyDomain.java
@@ -53,7 +53,7 @@ public class ModifyDomain extends AdminDocumentHandler {
           "can not access domain, domain is in " + domain.getDomainStatusAsString() + " status");
     }
 
-    checkDomainRight(zsc, domain, attrs);
+    AdminAccessControl adminAccessControl = checkDomainRight(zsc, domain, attrs);
 
     // check to see if domain default cos is being changed, need right on new cos
     checkCos(zsc, attrs, Provisioning.A_zimbraDomainDefaultCOSId);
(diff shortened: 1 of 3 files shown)


store/src/main/java/com/zimbra/cs/service/admin/ModifyDomain.java
```diff
@@ -94,12 +98,31 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+   * zimbraVirtualHostname. </Note>
+   *
+   * @param adminAccessControl {@link com.zimbra.cs.service.admin.AdminAccessControl}
+   * @param config the global config {@link com.zimbra.cs.account.Config}
+   * @return true if an authenticated user has right to modify global config, false otherwise.
+   * @throws ServiceException if something goes wrong during the checks.
+   * @author Yuliya Aheeva
+   * @since 23.6.0
+   */
+  protected boolean hasRight(AdminAccessControl adminAccessControl, Config config)
```
frisonisland, line 115: IMHO we can get along even without this method, because "hasRight" actually is "isAdmin and can modify global config"

keshavbhatt (approved):
lgtm.

store/src/test/java/com/zimbra/cs/service/admin/ModifyDomainIT.java
```diff
@@ -19,38 +20,65 @@
...
+import org.junit.After;
+import org.junit.Before;
+
 import org.junit.Rule;
 import org.junit.Test;
 import org.junit.rules.ExpectedException;
 
 public class ModifyDomainIT {
+  public static final String DELEGATED = "delegated";
+  public static final String WITH_RIGHT = "_with_right";
```
frisonisland, line 32: This test class could end up with many other tests, how do you ensure the account you create was not already created, thus making the test fail (potentially)? I would recommend treating each test as individual, with local variables. Consider using UUID for email local part of a user

store/src/main/java/com/zimbra/cs/service/admin/ModifyDomain.java
aheeva-yuliya: @frisonisland 
if it's a global admin then it won't even go deeper and check if he really has a right to modify global config bcz he has
if it's a delegated admin then it should go throuh all the chain and really check if he has this right bcz usually they don't have it
i didnt get your point how to get along without this method

keshavbhatt:
Please limit the length of the title of this PR. :D

store/src/test/java/com/zimbra/cs/service/admin/ModifyDomainIT.java
```diff
@@ -19,38 +20,65 @@
...
   }
 
   @Rule public ExpectedException expectedEx = ExpectedException.none();
 
-  private Account createDelegatedAdminForDomain(Domain domain) throws ServiceException {
+
+  private Account createAdminAccount(String attribute, String name) throws ServiceException {
     return provisioning.createAccount(
-        "admin@" + domain.getDomainName(),
+        name + "@" + DOMAIN_NAME,
```
frisonisland, line 70: I wouldn't use a static domain, to avoid conflict with users defined in other tests.
It may work now, but that is because the Provisioning class is not the real one used in production, it's the "Mock" which I found having many issues, such as not checking collision, etc.

frisonisland (changes requested):

store/src/main/java/com/zimbra/cs/service/admin/ModifyDomain.java
frisonisland: What I mean is the method is used just in one place, to me we could have written the condition inline, without defining this method. Also when will we ever use a method called "hasRight" in other places (in this class)?. The method name suggest a  generic check, and has generic parameters, but the behavior is very specific. 
I expect it to be either:
- "hasAdminRightAndCanModifyConfig)
-  hasRight(Right right)

aheeva-yuliya: @ frisonland i made it as a separate method in order to be able to override it in tests
if i would make it inline then tests for delegated admin will be gone
till now i change the name 
choose 😂 keep it inline or add tests

aheeva-yuliya: @frisonisland i would make it inline ok
but about tests 
i have to mock not an AccessManager but abstract AdminAccessControl or private static ACLAccessControl it's more natural but it's not possible
so removing tests for delegated admin

frisonisland: Ok, let's not remove tests :smile: 
Unfortunately there is a weird/confusing layout of the classes and no interface. We have to find a way to make this code more digestible while working on it somehow, by: make it non-static, use composition, etc.
For example: AdminDocumentHandler has a static method checkRight() that checks the right and returns current AccessControl instance from AccessManager.getInstance(). 
A solution could be, when creating ModifyDomain, passing AccessManager.getInstance() to the constructor. 
...maybe we can create another ticket on this.

aheeva-yuliya: @frisonisland noooooo I've already removed them 

i asked you to choose => you chose

What you want in the end? why you doing this to me on friday?

frisonisland: I just asked for the method to make sense :sob: I think I got confused, re-reading: no need to add tests for that method. 
I thought you were going to remove some tests from this comment

> but about tests
i have to mock not AccessManager but abstract AdminAccessControl or private static ACLAccessControl it's more natural but it's not possible (or link me how to do it)
so removing tests for delegated admin

frisonisland: Ok, keeping the method and not inline bc code is a mess and tests are more important.

Polpetta (approved):

Polpetta:
LGTM after @frisonisland's suggestions

frisonisland (approved):

PR #222, opened 2023-05-17
chore: upgrade slf4j and spring version

Polpetta:
CO-514


keshavbhatt (approved):

frisonisland:
@Polpetta you will hate me, but did we run a smoke test on an installation? Asking because slf4j libraries are also provided by this repo: https://github.com/Zextras/carbonio-zcs-lib
Some times ago we had Linking issue because of inconsistencies between mailbox dependencies and these.

store/pom.xml
```diff
@@ -408,7 +408,13 @@
...
     <dependency>
       <groupId>org.slf4j</groupId>
-      <artifactId>slf4j-log4j12</artifactId>
+      <artifactId>slf4j-reload4j</artifactId>
+      <scope>provided</scope>
+    </dependency>
+    <dependency>
+      <groupId>ch.qos.reload4j</groupId>
+      <artifactId>reload4j</artifactId>
+      <version>1.2.22</version>
```
frisonisland, line 417: These deps are provided by: https://github.com/Zextras/carbonio-zcs-lib
So we need to change this also.
These libraries are "common", they are not used only by mailbox but also Jetty and probably some other java application. Unfortunately some dependencies are shared.

frisonisland (changes requested):

Polpetta:
Attached to this PR is also https://github.com/Zextras/carbonio-zcs-lib/pull/18 now

store/pom.xml
frisonisland: I'm not sure we need slf4j-reload4j, as slf4j-api is still available: https://mvnrepository.com/artifact/org.slf4j/slf4j-api
Seems reload4j is a drop-in replacement for log4j 1.2.17 (https://github.com/qos-ch/reload4j) so very different from log4j 2

frisonisland:
This PR may be partially obsolete after #237 . Still we have to bump Spring, but it will be easier since we will merge zcs-lib in this repo and will only have to update 1 pom if we are lucky (dependency management)

keshavbhatt:
@zextras/nbm can we close this as the migration is completed already?

Polpetta:
I'll close it, if someone else this this makes sense to merge, please reopen it

PR #233, opened 2023-05-22
chore: performance enhancements applying various language level fixes

keshavbhatt:
**What has changed:**
- avoid new Boolean
- bulk method call instead of manually operating with loops
- avoid passing fix array size in Collection.toArray calls
- add break in loops when intended condition matches
- replace manual array copy with System.arraycopy
- replace unnecessary calls to Collection.addAll
- replace unnecessary calls to String.format
- replace unnecessary manual array to collection copy
- replace string concatenation in loop with StringBuilder/StringBuffer.append to prevent unnecessary copying of string
- replace string concatenation with StringBuilder/StringBuffer.append to prevent unnecessary allocations of new Objects
- replace redundant PrintStream.printf calls with PrintStream.print
- replace loop which are doing string concat with StringBuffer/Builder/Appendable.append with String.repeat

**Tests:** 
- Existing unit tests passes
- Smoke test passes

**Note:** pre-commit hooks were avoided and will be applied as single commit at the end.


client/src/main/java/com/zimbra/client/ZInvite.java
```diff
@@ -1600,7 +1600,7 @@ public ZJSONObject toZJSONObject() throws JSONException {
         }
 
         public String toString() {
-            return String.format("[ZRecurrenceDate]"); // TODO
```
frisonisland, line 1603: What are all these TODO?

keshavbhatt: mystery :crying_cat_face:

store/src/main/java/com/zimbra/cs/wiki/WikiDigestFixup.java
```diff
@@ -134,7 +134,7 @@ private static List<WikiDigest> getWikiDigests(int mboxId) throws IOException, S
     private static void fixupItems(DbConnection conn, int mboxId, List<WikiDigest> digests)
     throws SQLException, ServiceException {
         StringBuilder sql = new StringBuilder("UPDATE ");
-        sql.append("mailbox" + mboxId + ".mail_item");
```
frisonisland, line 137: Please check for this class usages, Wiki should have been removed in https://github.com/Zextras/carbonio-mailbox/pull/217 I think I missed this

keshavbhatt: Yes, I am aware of it and i was about to ping you for this. Will remove it now since you confirmed. There is also one more class called PageCache under same package.

frisonisland: Thanks, I saw there are still many Wiki usages, I vote for cleaning it properly in another future PR

keshavbhatt: Database cleanup should also be performed if not done already.

frisonisland:
Generally speaking it looks good to me. I would choose _println_ over using _printf_ with new line character, but that's ok.
Well, actually I have a strong opinion: to not use System.out but a log instance instead :smile: , but we will work on that

keshavbhatt:
> I have a strong opinion: to not use System.out but a log instance instead

We will work on this in future. These issues are tracked by SC atm.

frisonisland (approved):

aheeva-yuliya (approved):

store/src/main/java/com/zimbra/cs/imap/ImapRequest.java
```diff
@@ -1023,14 +1024,15 @@ protected ImapPartSpecifier readPartSpecifier(boolean binary, boolean literals)
                 }
                 skipChar(')');
             } else if (sectionText.equals("MIME")) {
-                if (sectionPart.isEmpty()) {
+                if (sectionPart.length() == 0) {
```
Polpetta, line 1027: According to Java implementation of `isEmpty()`:
```java
    public boolean isEmpty() {
        return this.value.length == 0;
    }
```

I think that `length() == 0` and `isEmpty()` are equals in term of performance and readability of the codebase. If you want to improve the robustness of this check, what you could do is instead:

```suggestion
                if (sectionPart.isBlank()) {
```

This filters empty spaces too.

store/src/main/java/com/zimbra/cs/mailbox/DeliveryOptions.java
```diff
@@ -53,7 +53,7 @@ public DeliveryOptions setFlags(int flags) {
     }
 
     public DeliveryOptions setTags(Collection<String> tags) {
-        mTags = tags == null ? null : tags.toArray(new String[tags.size()]);
+        mTags = tags == null ? null : tags.toArray(new String[0]);
```
Polpetta, line 56: Since you're here, please convert this ternary operator to a standard `if` statement, or use `Optional` if you prefer to improve readability.

Polpetta (changes requested):
A part from some of the suggestions I made, I have some doubts about this PR. In particular:
* while these changes are appreciated, they are modifying a lot of files. With other big PRs WIP this will generate some conflicts while merging and I don't know if the improvements are worth all the conflicts. If we go and merge it anyway, I'd give it low priority in order to avoid making difficult to merge other PRs.
* Java compiler will automatically rewrite short `+` concatenations starting from Java compiler 9+ under the hood with `StringBuffer` or `StringBuilder` accordingly. I think `+` is more readable than `.append()`, and in any case Google Java Formatter splits new lines with `+`, so we'll end up with a mix of the twos. Real improvement should be using `.append()` while in loops and other iterative methods (https://www.baeldung.com/java-string-performance).
* I still see in the codebase that we're not marking variables that do not change reference as `final`. I think it should be included in this PRs too if we decide to merge it. Marking variables as `final` will improve codebase soundness and allows the compiler to make optimizations based on the assumption that the memory reference will not change for that variable lifetime, especially for codebase that is multi-thread and that the compiler cannot demonstrate at compile time that the reference will not change. _This is a very important codebase improvement and optimization we can make and that we miss every time we merge variables without marking them as final_. In IntelliJ there is a specific checkbox to mark in order to create variables as `final` by default.

Finally, while (I couldn't stress it enough) I appreciate these changes to improve the overall quality, we're still improving milliseconds or half seconds on a system that takes seconds to perform SOAP API response because of big bottlenecks on the system (queries for instance).

client/src/main/java/com/zimbra/client/ZInvite.java
Polpetta: I think we can remove that TODO and mark it as """resolved""" :joy:

keshavbhatt:
> * PRs WIP this will generate

I saw no big PRs, and thats why i choose to work on this in the begining of the new sprint.



> * Java compiler will automatically rewrite short `+` concatenations starting from Java compiler 9+

I agree, But if you will see we are also replacing `+` in stringbuilder case's and in loops. Also, i did not performed these changes manually with my hands :D i used intellij static analysis for mass refactoring. So, all the changes could not be directly impacting performance offcourse. 



> * I still see in the codebase that we're not marking variables that do not change reference as `final`

This is something i have planned for another PR. I am working on this and manay more imporvements. 

>I appreciate these changes to improve the overall quality we're still improving milliseconds or half seconds on a system that takes seconds to perform SOAP API response

Thanks for the appreciation :) . I did not performed any benchmarking on these changes but, i am confidant that these changes will improve by tiny fraction. Also, most of the SOAP calls are not taking seconds. Even on the the testing VMs we use(which are very low on resources) most of the SOAP calls usually takes milliseconds to get processed.

store/src/main/java/com/zimbra/cs/mailbox/DeliveryOptions.java
keshavbhatt: [43399fe](https://github.com/Zextras/carbonio-mailbox/pull/233/commits/43399fec792f7422e63c1fbac1d8a87af4561ad8) fixes this.

store/src/main/java/com/zimbra/cs/imap/ImapRequest.java
keshavbhatt: `sectionPart` is StringBuilder which has no method `isBlank`.

Polpetta: Then it's ok

Polpetta (approved):

keshavbhatt:
Not formatting touched files, as it will make conflict resolution hard for few WIP PRs.

PR #245, opened 2023-06-06
feat: [CO-787] add copy to files with shared mailbox

frisonisland:
These changes add the ability to retrieve an attachment from another mailbox, if specified in the request.
Similarly yo UserServlet, the messageId it's not actually an id but can be in form of <UUID:id> where UUID represents another account/mailbox.

- add test to check shared mailbox is used to search for attachment
- split messageId in UUID:messageId


Polpetta (approved):

frisonisland:
Closing as requires more testing, also the client behaves very differently in prod vs development due to underlying implementation of InputStream, using https://github.com/Zextras/carbonio-mailbox/blob/2eb731841c18fb0535a0bd0b49b61dca989e91a1/common/src/main/java/com/zimbra/common/zmime/ZTransferEncoding.java#L39
First goal is to make UT and IT work. Found E2E are failing, so need to re-evaluate underlying implementations of InputStream

PR #247, opened 2023-06-07
chore: use log4j2 only

frisonisland:
**Edit:** merged devel into this, that is why there are many changes

**Changes:**
- updated poms, no log4j1.X
  - zm-store and store-extra-runtime-dependencies use log4j2 (log4j-core and log4j-api), so I declared it in respective poms.
  - Note: from an analysis zookeeper dependency depends on old log4j, so it is still pulled in, but we do not use it
```
  +- com.101tec:zkclient:jar:0.11:compile
[INFO] |  +- org.apache.zookeeper:zookeeper:jar:3.4.5:compile
[INFO] |  |  +- org.slf4j:slf4j-log4j12:jar:1.7.30:compile
[INFO] |  |  +- log4j:log4j:jar:1.2.17:compile
[INFO] |  |  +- jline:jline:jar:0.9.93:compile
[INFO] |  |  |  \- junit:junit:jar:4.8.2:compile
[INFO] |  |  \- org.jboss.netty:netty:jar:3.2.2.Final:compile
[INFO] |  \- org.slf4j:slf4j-api:jar:1.7.30:compile
```

- updated .properties files
- updated log4j 1.X code usages (org.apache.log4j -> org.apache.logging.log4j)
- removed Log4J2ConfigurationFactory. Reason is **when requesting a logger, log4j calls Configuration Factory**. This creates a Factory instance + loads factories marked as plugins. We can pass **"log4j.configurationFactory" and "log4j.configurationFile"** to jvm (log4j.configurationFactory is not needed if the custom ConfigurationFactory is marked as plugin)
The **first one alone doesn't work, because it expects also a configurationFile**, else it fallbacks to DefaultConfiguration.
Passing only **log4j.configurationFIle to JVM makes log4j use default Factory with .properties file**. Since **Log4j2ConfigurationFactory has a pretty standard logic, I think we can drop it.**


frisonisland:
I have some doubts about static usage of Configurator, I'm not very fond of it.

Polpetta (approved):

keshavbhatt (approved):

PR #249, opened 2023-06-08
feat: [CO-721] delete Certbot domain config

aheeva-yuliya:


store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -2360,6 +2362,55 @@ private static void deleteObsoleteCertificateKeyPairFromDomainCertDir(
...
+  /**
+   * Executes Certbot delete command.
+   *
+   * @param args domain names to perform Certbot cleanup for
+   * @throws ProxyConfException in case of failure during the deletion
+   * @author Yuliya Aheeva
+   * @since 23.7.0
+   */
+  private static void executeCertbotDelete(final List<String> args) throws ProxyConfException {
+    args.forEach(domainName -> LOG.info("Deleting Let's Encrypt configuration for " + domainName));
```
keshavbhatt, line 2402: What are we doing here? 

Args contains domain names, we log that we are "Deleting Let's Encrypt configuration for each domain item",
Then we updating args with `certbot` path and `delete` argument, and executing process builder with args? 

Can we make it little simple? 
By looping these steps instead:
- log deletion
- delete
- log if deletion failed

For example:
```
loop
    LOG.info("Deleting Let's Encrypt configuration for " + domainName);
    List<String> deleteArgs = new ArrayList<>();
    deleteArgs.add(CERTBOT);
    deleteArgs.add("delete");
    deleteArgs.add(domainName);
    try
        new ProcessBuilder(deleteArgs).start();
    catch
        LOG.info("Something went wrong while deleting Let's Encrypt configuration for " + domainName);
```

keshavbhatt (changes requested):
added a suggestion.

frisonisland (approved):

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
aheeva-yuliya: idk why we should start new proccess for each domain in a loop
it's easier in my point of view to start one for all of them

keshavbhatt: Oh I just saw the certbot wrapper is handling bulk requests [here](https://github.com/Zextras/carbonio-core-utils/pull/48/files#diff-fda67076b2b22c16c5f4aca1be6839426dee28662d18edeec6c9f1721bb3ae1fR16).

keshavbhatt (approved):

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
aheeva-yuliya: @keshavbhatt is it ok to proceed this way then in your opinion?

Polpetta (approved):

PR #251, opened 2023-06-12
fix(CO-744): add back ChildAccount

frisonisland:
Adding back because AttributeManager produces a nasty warning when using zmprov cli

### DIFF
--- a/store/src/main/java/com/zimbra/cs/account/callback/ChildAccount.java
+++ b/store/src/main/java/com/zimbra/cs/account/callback/ChildAccount.java
@@ -16,7 +16,11 @@ import com.zimbra.cs.account.AttributeCallback;
 import com.zimbra.cs.account.Entry;
 import com.zimbra.cs.account.Provisioning;
 import com.zimbra.common.account.Key.AccountBy;
- 
+
+/**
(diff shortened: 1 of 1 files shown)


store/src/main/java/com/zimbra/cs/account/callback/ChildAccount.java
```diff
@@ -0,0 +1,97 @@
...
+
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.account.Account;
+import com.zimbra.cs.account.AccountServiceException;
+import com.zimbra.cs.account.AttributeCallback;
+import com.zimbra.cs.account.Entry;
+import com.zimbra.cs.account.Provisioning;
+import com.zimbra.common.account.Key.AccountBy;
+ 
+public class ChildAccount extends AttributeCallback {
```
Polpetta, line 20: At this point, @frisonisland mark this class and everything inside as deprecated. If you can, add why we cannot remove it as Javadoc inside the `@deprecated` tag

Polpetta (approved):
It's ok for me, just adding 2 cents

aheeva-yuliya (approved):

keshavbhatt (approved):

PR #253, opened 2023-06-13
fix: [CO-740] check retention policy before setting

keshavbhatt:
**What has changed:**
- Setting retention policy fails if both(keep and purge) policies are specified
- Setting retention policy fails if no policy is specified
- Unit test covering both cases


keshavbhatt:
For now keep option was removeed from the UI. Cleanup is planned for future.

PR #254, opened 2023-06-15
chore: junit5 migration

frisonisland:
These changes add JUnit5 and remove old Junit4 and Junit3 tests.
Also it cleans up some unused methods in some test utility classes

Also:

- XMLUnit has been migrated (1.X -> 2.X) for compatibility
- JLine dependency (a console writer) has excluded JUnit 3.8 (mistery why it brings in JUnit)
- Expected Exception checks have been migrated to assertThrows
- Parameterized has been removed in favor of new Parameterized syntax, that can be method-specific
- @Rule and some watcher classes, used to print debug information, have been removed
- TestUtil has been cleaned up from unused methods.
- Added test execution order in MigrateAttributeTest (test fails if executed in a specific order, even in devel branch, very concerning)
- Fixup FileLogWriter test to use new Junit5 temporary directory
- Fixup CreateMessageTest to assert on arrayEquals instead of assertEquals (was comparing objects instead of values)
- **Updated test stage name in Jenkinsfile** to a more appropriate one


Polpetta (approved):
I did a quick review via IntelliJ and it seems good to me! Let's try to bump this library :+1:

keshavbhatt (approved):
lgtm.

PR #255, opened 2023-06-16
feat(CO-727): allow Save To Files from a shared mailbox

frisonisland:
This PR expands the Save-To-Files by enabling the feature from a shared mailbox.

Changes:
- Allow save to files with mid <UUID:id>, where UUID is other account  UUID, id is message id
- **Add MockServer** (https://www.mock-server.com/) to mock Files (this spots runtime exception on files-client in case of different dependencies version)
- **Update httpmime** 4.5.13 -> 4.5.14 (https://mvnrepository.com/artifact/org.apache.httpcomponents/httpmime)
- **Update httpcore** 4.4.14 -> 4.4.16 (https://mvnrepository.com/artifact/org.apache.httpcomponents/httpcore)
- Add test for shared mailbox
- Add parameterized test for different attachment (mime can have different behavior in size)
- **Use Try.withResources to close the stream** when used (this was missing in previous implementation)
- Add INTERNAL_ERROR in ServiceException


store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -138,26 +128,31 @@ private Try<CopyToFilesRequest> getRequestObject(Element request) {
...
+    } else {
+      accountUUID = context.getAuthtokenAccountId();
+      msgId = request.getMessageId();
+    }
     final Try<Integer> messageIdTry =
-        Try.of(() -> Integer.parseInt(request.getMessageId()))
+        Try.of(() -> Integer.parseInt(msgId))
             .onFailure(ex -> mLog.debug(ex.getMessage()))
             .mapFailure(
                 Case(
```
keshavbhatt, line 146: How about using something cleaner like recoverWith instead of mapFailure:

```
.recoverWith(ex->Try.failure(ServiceException.PARSE_ERROR(MailConstants.A_MESSAGE_ID + " must be an integer.", ex)))```

File: store/src/test/java/com/zimbra/cs/service/mail/CopyToFilesIT.java [hunk 2]
```diff
@@ -32,271 +44,347 @@
...
 import org.junit.jupiter.api.Test;
-
+import org.junit.jupiter.params.ParameterizedTest;
+import org.junit.jupiter.params.provider.Arguments;
+import org.junit.jupiter.params.provider.MethodSource;
+import org.mockserver.integration.ClientAndServer;
+import org.mockserver.model.HttpResponse;
 
 /** Integration tests for CopyToFiles */
 public class CopyToFilesIT {
```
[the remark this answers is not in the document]
frisonisland: Yes, thx for pointing it out

store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
frisonisland: I didn't use recoverWith because the name implies a recover from an error and returning a failure doesn't seem right, because it is like recovering with a failure.
The initial idea was to make sure no other exception aside ServiceException is thrown, as it is a managed exception, that's why the mapFailure.
However I see now it maps "Exception" which is too generic, so even mapFailure doesn't make much sense. I will check and propose another solution for this.

frisonisland: I agree recoverWith is cleaner to read in first case. I see there are many exceptions that could happen and the mapping may be hiding other errors. For example "getAttachmentToCopy" always returns "messageId must be an integer", which is wrong.

frisonisland: Decided to leave mapFailure on this, as I've added two cases: numberFormatException and Exception.
In the former case maps to a specific parse error, else internal error. + added tests on this.
Let me know!

store/src/test/java/com/zimbra/cs/service/mail/CopyToFilesIT.java
```diff
@@ -32,271 +44,347 @@
...
   @BeforeEach
   public void setUp() throws Exception {
     MailboxTestUtil.initServer();
+    filesServer = startClientAndServer(20002);
     Provisioning prov = Provisioning.getInstance();
+    final Account sharedAcct =
+        prov.createAccount(
+            "shared@zimbra.com",
+            "secret",
+            new HashMap<String, Object>() {
```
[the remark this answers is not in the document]
frisonisland: fixed

store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
```diff
@@ -53,65 +54,57 @@ public CopyToFiles(AttachmentService attachmentService, FilesClient filesClient)
...
-    // get attachment name
-    final Try<String> fileNameTry = attachmentTry.flatMapTry(this::getAttachmentName);
-    // get attachment content stream
-    final Try<InputStream> attachmentStreamTry =
-        attachmentTry.mapTry(Part::getInputStream).onFailure(ex -> mLog.debug(ex.getMessage()));
-
-    // execute Files api call and flatten yield result
-    // maps every non-ServiceException to internal error
-    String nodeId =
+    final NodeId nodeId =
```
keshavbhatt, line 57: This whole chain of logic scares me and look ugly but i will pass since it will take a lot of refactoring :/

store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
frisonisland: In ServiceException.INTERNAL_ERROR we are providing the exception. You mean returning the exception in response?

keshavbhatt (approved):
LGTM.

I don't recommend the current exception propagation approach used in this codebase because it makes it difficult to track them in the calling function. I think it can be suitable in certain scenarios where the exceptions are not critical and can be disregarded.

Additionally, I suggest avoiding excessive use of functional chaining as it makes the code appear messy and unorganized.

store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
frisonisland: You know, I kinda agree, but you'll have to fight with @Polpetta :laughing: 
The issue is we are using a lot of methods with trys, that is why multiple flatmaps and yield.
Also there is the issue of using an inputstream, which requires closing it.

IMHO we have to decide together some basic rules and tools so we avoid conflicting opinions on flow and styling.
Also for minor things like "use import static" (not everyone agrees: https://docs.oracle.com/javase/8/docs/technotes/guides/language/static-import.html it is a team preference)

Bottom line: if we want ABSOLUTELY to use vavr, then I'm afraid this is the way

store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
keshavbhatt: I mean the "intenal error" seems too wide to understand what went wrong. My doubt is are we returning the actual message of real exception or not? having that message logged or returned will be hepful rather than just saying "internal error".

frisonisland: No, we are not returning it, because I am not sure we want to do it. I think it should be logged.
I won't say it is right or wrong, rather a common lack of design among APIs.

keshavbhatt: :heavy_check_mark: It should be logged.

frisonisland:
> LGTM.
> 
> I don't recommend the current exception propagation approach used in this codebase because it makes it difficult to track them in the calling function. I think it can be suitable in certain scenarios where the exceptions are not critical and can be disregarded.
> 
> Additionally, I suggest avoiding excessive use of functional chaining as it makes the code appear messy and unorganized.

I see this class as a **Controller**, that's why I decided to map the failure on each validation step (check token, check body, check messageId, check file exists, etc.). Then all is **chained in the handle** method
Unfortunately **about functional** I cannot say much, some people like it more, some not (aside other technical benefits). 
This is also how vavr works.
I think **we should opt for a common decision that works for us**.

store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
frisonisland: The underlying error is logged at debug level before mapping each failure, on each method. I think it's better to log at ERROR level, agree?

store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
Polpetta: I like a functional programming approach, so I like this kind of chains because I think they're more declarative rather than a set of imperative statements or a collection of objects working together. In this case it is unfortunate that there are so many `Try`. We should rather ask if we set the `Try` in the right place along the chain of calls.

I don't mind using vavr, it has been around for quite some time and the codebase is relatively small. It is invasive, but because it is filling holes in the jdk basically.

I think that the next approach for the mailbox is a stateless, functional like one, and since we have Java as language...when life give you lemons, make lemonade

Polpetta:
> Additionally, I suggest avoiding excessive use of functional chaining as it makes the code appear messy and unorganized.

Sorry, but I'm completely opposite to this statement. Please do use more functional chaining as it makes the code more maintainable by making it declarative

Functional seems scary at the beginning, but once it enters in your mind and you understand how it works it is a breeze to your programming experience.

keshavbhatt:
> Sorry, but I'm completely opposite to this statement.

I dont have issues with employing funtional apparoach to write the code. The issue I was trying to point out is "excessive use of functional chaining" that we can see in this code. That is making the code hard to debug and read. 

> "The code appear messy and unorganized" 

I mean :arrow_heading_down: 

 ```
   final NodeId nodeId =
        Optional.ofNullable(
                getRequestObject(request)
                    .flatMap(
                        copyToFilesRequest ->
                            this.getAttachmentToCopy(copyToFilesRequest, zsc)
                                .flatMap(
                                    attachment ->
                                        Try.withResources(attachment::getInputStream)
                                            .of(
                                                inputStream ->
                                                    For(
                                                            Try.of(
                                                                () ->
                                                                    ZimbraCookie
                                                                            .COOKIE_ZM_AUTH_TOKEN
                                                                        + "="
                                                                        + zsc.getAuthToken()
                                                                            .getEncoded()),
                                                            this.getDestinationFolderId(
                                                                copyToFilesRequest),
                                                            this.getAttachmentName(attachment),
                                                            this.getAttachmentContentType(
                                                                attachment),
                                                            Try.withResources(
                                                                    attachment::getInputStream)
                                                                .of(this::getAttachmentSize)
                                                                .flatMap(identity()))
                                                        .yield(
                                                            (cookie,
                                                                folderId,
                                                                fileName,
                                                                contentType,
                                                                fileSize) ->
                                                                filesClient.uploadFile(
                                                                    cookie,
                                                                    folderId,
                                                                    fileName,
                                                                    contentType,
                                                                    inputStream,
                                                                    fileSize)))))
                    .flatMap(identity())
                    .flatMap(identity())
                    .mapFailure(
                        Case(
                            $(ex -> !(ex instanceof ServiceException)),
                            ServiceException::INTERNAL_ERROR))
                    .get())
            .orElseThrow(() -> ServiceException.FAILURE("got null response from Files server."));
```

**Comment by frisonisland** [comment]
> > Sorry, but I'm completely opposite to this statement.
> 
> I dont have issues with employing funtional apparoach to write the code. The issue I was trying to point out is "excessive use of functional chaining" that we can see in this code. That is making the code hard to debug and read.
> 
> > "The code appear messy and unorganized"
> 
> I mean arrow_heading_down
> 
> ```
>   final NodeId nodeId =
>        Optional.ofNullable(
>                getRequestObject(request)
>                    .flatMap(
>                        copyToFilesRequest ->
>                            this.getAttachmentToCopy(copyToFilesRequest, zsc)
>                                .flatMap(
>                                    attachment ->
>                                        Try.withResources(attachment::getInputStream)
>                                            .of(
>                                                inputStream ->
>                                                    For(
>                                                            Try.of(
>                                                                () ->
>                                                                    ZimbraCookie
>                                                                            .COOKIE_ZM_AUTH_TOKEN
>                                                                        + "="
>                                                                        + zsc.getAuthToken()
>                                                                            .getEncoded()),
>                                                            this.getDestinationFolderId(
>                                                                copyToFilesRequest),
>                                                            this.getAttachmentName(attachment),
>                                                            this.getAttachmentContentType(
>                                                                attachment),
>                                                            Try.withResources(
>                                                                    attachment::getInputStream)
>                                                                .of(this::getAttachmentSize)
>                                                                .flatMap(identity()))
>                                                        .yield(
>                                                            (cookie,
>                                                                folderId,
>                                                                fileName,
>                                                                contentType,
>                                                                fileSize) ->
>                                                                filesClient.uploadFile(
>                                                                    cookie,
>                                                                    folderId,
>                                                                    fileName,
>                                                                    contentType,
>                                                                    inputStream,
>                                                                    fileSize)))))
>                    .flatMap(identity())
>                    .flatMap(identity())
>                    .mapFailure(
>                        Case(
>                            $(ex -> !(ex instanceof ServiceException)),
>                            ServiceException::INTERNAL_ERROR))
>                    .get())
>            .orElseThrow(() -> ServiceException.FAILURE("got null response from Files server."));
> ```

TBH it's just the "For" that makes the code messy, but the rest is quite straightforward

**Comment by frisonisland** [comment]
> > "The code appear messy and unorganized"
> 
> I mean arrow_heading_down
> 
> ```
>   final NodeId nodeId =
>        Optional.ofNullable(
>                getRequestObject(request)
>                    .flatMap(
>                        copyToFilesRequest ->
>                            this.getAttachmentToCopy(copyToFilesRequest, zsc)
>                                .flatMap(
>                                    attachment ->
>                                        Try.withResources(attachment::getInputStream)
>                                            .of(
>                                                inputStream ->
>                                                    For(
>                                                            Try.of(
>                                                                () ->
>                                                                    ZimbraCookie
>                                                                            .COOKIE_ZM_AUTH_TOKEN
>                                                                        + "="
>                                                                        + zsc.getAuthToken()
>                                                                            .getEncoded()),
>                                                            this.getDestinationFolderId(
>                                                                copyToFilesRequest),
>                                                            this.getAttachmentName(attachment),
>                                                            this.getAttachmentContentType(
>                                                                attachment),
>                                                            Try.withResources(
>                                                                    attachment::getInputStream)
>                                                                .of(this::getAttachmentSize)
>                                                                .flatMap(identity()))
>                                                        .yield(
>                                                            (cookie,
>                                                                folderId,
>                                                                fileName,
>                                                                contentType,
>                                                                fileSize) ->
>                                                                filesClient.uploadFile(
>                                                                    cookie,
>                                                                    folderId,
>                                                                    fileName,
>                                                                    contentType,
>                                                                    inputStream,
>                                                                    fileSize)))))
>                    .flatMap(identity())
>                    .flatMap(identity())
>                    .mapFailure(
>                        Case(
>                            $(ex -> !(ex instanceof ServiceException)),
>                            ServiceException::INTERNAL_ERROR))
>                    .get())
>            .orElseThrow(() -> ServiceException.FAILURE("got null response from Files server."));
> ```

@Keshav I can do the following (introducing a method to get the cookie):
```
Optional.ofNullable(
                getRequestObject(request)
                    .flatMap(
                        copyToFilesRequest ->
                            this.getAttachmentToCopy(copyToFilesRequest, zsc)
                                .flatMap(
                                    attachment ->
                                        Try.withResources(attachment::getInputStream)
                                            .of(
                                                inputStream ->
                                                    For(this.getCookie(zsc),
                                                            this.getDestinationFolderId(copyToFilesRequest),
                                                            this.getAttachmentName(attachment),
                                                            this.getAttachmentContentType(attachment),
                                                            Try.of(() -> inputStream),
                                                            Try.withResources(
                                                                    attachment::getInputStream)
                                                                .of(this::getAttachmentSize)
                                                                .flatMap(identity()))
                                                        .yield(filesClient::uploadFile))))
                    .flatMap(identity())
                    .flatMap(identity())
                    .onFailure(ex -> mLog.error(ex.getMessage()))
                    .mapFailure(
                        Case(
                            $(ex -> !(ex instanceof ServiceException)),
                            ServiceException::INTERNAL_ERROR))
                    .get())
                    .orElseThrow(() -> ServiceException.FAILURE("got null response from Files server."));
```

**Comment by Polpetta** [comment]
The problem is not the functional approach, is that there are a lot of objects to unwrap. This is kinda a bad design for whom implemented it in the past, little we can do with such a codebase other than trying our best.
The `getCookie` helps a little I agree

File: store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java [hunk 5]
```diff
@@ -53,65 +53,38 @@ public CopyToFiles(AttachmentService attachmentService, FilesClient filesClient)
...
+                                                                attachment),
+                                                            Try.of(() -> inputStream),
+                                                            this.getAttachmentSize(attachment))
+                                                        .yield(filesClient::uploadFile))))
+                    .flatMap(identity())
+                    .flatMap(identity())
+                    .onFailure(ex -> mLog.error(ex.getMessage()))
                     .mapFailure(
                         Case(
                             $(ex -> !(ex instanceof ServiceException)),
```
Polpetta, line 82: I know this is not a change of this pr, but please check this statement, there should be something like `instanceof()` method instead of the raw call. See if you can apply it

store/src/test/resources/test-save-to-files.jpg
Polpetta, line 1: While I like this image and I totally approve it, is it copyright free?

Polpetta:
Also I saw that you edited stuff related to certbot in `ProxtConfGen`, like `deleteDomainCertbotConfiguration`. Why?

store/src/test/resources/test-save-to-files.jpg
frisonisland: Good question

frisonisland: Fixed

frisonisland:
> Also I saw that you edited stuff related to certbot in `ProxtConfGen`, like `deleteDomainCertbotConfiguration`. Why?

Check again, there are no changes on ProxyConfGen!

keshavbhatt (approved):

store/src/main/java/com/zimbra/cs/service/mail/CopyToFiles.java
Polpetta: Great!

store/src/test/resources/test-save-to-files.jpg
Polpetta: Ty, sorry for being an hassle

Polpetta (approved):
LGTM

store/src/test/resources/test-save-to-files.jpg
frisonisland: No you did right checking, I was being lazy :sweat_smile:

PR #265, opened 2023-07-07
chore: upgrade pre-commit to new common rules

Polpetta:
Basically with @federicorispo and @nomanAli95 on behalf of the backend tribe we decided new common rules for `pre-commit` hooks. With this PR, we update the mailbox to these new rules (a document is being prepared with pre-defined templates to copy from, this is basically one of them)

I didn't activate reuse because it runs on all the codebase and makes the build fail. We have to fix our headers before being able to activate it!

### DIFF
--- a/.markdownlint.yaml
+++ b/.markdownlint.yaml
@@ -1,0 +1,4 @@
+default: true
+MD013:
+  line_length: 120
+  code_blocks: false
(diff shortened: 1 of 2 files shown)


frisonisland (approved):
+1 for commit-msg

keshavbhatt (approved):

PR #266, opened 2023-07-07
feat: [CO-767] Remove stale config from Carbonio SOAP Service config 

keshavbhatt:
- all zimlet related config was removed
- change the display name of service to Carbonio SOAP Service


store/conf/web.xml
```diff
@@ -82,17 +82,6 @@ SPDX-License-Identifier: AGPL-3.0-only
     <url-pattern>/*</url-pattern>
   </filter-mapping>
 
-  <filter>
-    <filter-name>ThrottlingFilter</filter-name>
-    <filter-class>com.zimbra.cs.servlet.ThrottlingFilter</filter-class>
```
frisonisland, line 87: What about removing the class completely?

frisonisland (commented):
Just want to know if we can remove also zimlet Filter classes

keshavbhatt:
> Just want to know if we can remove also zimlet Filter classes

We will cover that in CO-778

store/conf/web.xml
keshavbhatt: We will cover that in CO-778

frisonisland (approved):

Polpetta (approved):

PR #267, opened 2023-07-11
fix(CO-782): Delete Draft from authed user's mailbox in SendMsg

frisonisland:
Delete draft always from authed user's mailbox when sending an email.

This because when sending an email with a shared mailbox, the Draft is saved by the client on the authed user's mailbox.
It is a temporary fix for the **issue of not deleting the draft when acting as shared account** until we come up with a better flow.

Changes:
- bump to 23.7.1 (PKGBUILD + CHANGELOG in another PR to not lose commit after squash)
- fix behavior + integration test with GreenMail
  - test: draft deleted from authed user's mailbox
  - test: draft not deleted from shared user's mailbox

### DIFF
--- a/ant-build-common.xml
+++ b/ant-build-common.xml
@@ -63,7 +63,7 @@
   </target>
 
   <target name="carbonio-version">
-    <property name="carbonio.buildinfo.version" value="23.7.0_ZEXTRAS_202307"/>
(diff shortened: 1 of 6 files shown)


keshavbhatt (approved):
LGTM.

Polpetta (approved):

PR #276, opened 2023-07-12
chore: update the package verions to match the expected one

keshavbhatt:
- carbonio-appserver-store-libs
- carbonio-common-core-libs

4.0.10 was put mistakenly while merging carbonio-zcs-lib in mailbox. 
This change will make above mentioned mailbox pakcages install latest changes for us everywhere.

### DIFF
--- a/packages/appserver-store-libs/PKGBUILD
+++ b/packages/appserver-store-libs/PKGBUILD
@@ -3,7 +3,7 @@ targets=(
   "ubuntu"
 )
 pkgname="carbonio-appserver-store-libs"
-pkgver="4.0.10"
+pkgver="4.0.11"
 pkgrel="1"
 pkgdesc="Replace Carbonio store libs"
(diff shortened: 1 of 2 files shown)


packages/appserver-store-libs/PKGBUILD
```diff
@@ -3,7 +3,7 @@ targets=(
   "ubuntu"
 )
 pkgname="carbonio-appserver-store-libs"
-pkgver="4.0.10"
```
frisonisland, line 6: I think we have to set it at 4.6.1, same as other packages.
The bump will not be an issue, but release-it will be able to manage it

keshavbhatt: Oh, Okay. I thought we would do that while release. I will udpate.

Polpetta (approved):

PR #287, opened 2023-07-19
chore: [CO-791] remove Tasks related code

aheeva-yuliya:


Polpetta (changes requested):
Everything seems ok, just remove that extra comment that is not necessary imho

frisonisland (approved):

keshavbhatt:
@aheeva-yuliya 

We can also have some cleanup on the MailItemType: https://github.com/zextras/carbonio-mailbox/blob/chore_CO-791_remove_task_feature/common/src/main/java/com/zimbra/common/mailbox/MailItemType.java#L35

Please remove Note type (has no usages) which I missed in one of my recent PR. 
And also other which have no usages if you want.

Polpetta (approved):

PR #292, opened 2023-07-25
fix: [CO-803] Fix NPE while exporting user data

keshavbhatt:
when user have no Calendar items in their account.

**What has changed:**
- StringBuilder object used to accumulate some "extra" user data is initialized instead of being assigned to null avoiding NPE in successive calls.


keshavbhatt:
Now provided by #293

PR #299, opened 2023-08-02
fix: [CO-814] jython downgrade to 2.5.2 in order to fix configd

M0Rf30:


frisonisland:
Jython is not used directly by this codebase, but only by zmconfigd. What if we move this dependency to somewhere else or zmconfigd repo? Considering also zm-jython-standalone weighs a lot.
Also by downgrading we will bring in a vulnerability: https://mvnrepository.com/artifact/org.python/jython-standalone

M0Rf30:
No problems on my side. I think that your effort will take longer. This is to fix the issue now.
@flanciotti what do you think?

frisonisland (changes requested):
Before proceeding we should:
- provide description why we need this change
- explain the need of the downgrade, since jython was not bumped in recent versions of Carbonio
- provide some tests to support the change

Also the downgrade will lessen the quality of the product, as it will bring this critical CVE: https://nvd.nist.gov/vuln/detail/CVE-2016-4000

mgarbin:
> Before proceeding we should:
> 
>     * provide description why we need this change
> 
>     * explain the need of the downgrade, since jython was not bumped in recent versions of Carbonio
> 
>     * provide some tests to support the change
> 
> 
> Also the downgrade will lessen the quality of the product, as it will bring this critical CVE: https://nvd.nist.gov/vuln/detail/CVE-2016-4000

Inside internal ticket is already written all you need, i tagged you.

frisonisland (approved):
Jython was upgraded in: https://github.com/zextras/carbonio-zcs-lib/commit/cb8cda9d8b3fb529d762d5dc21567cd1035729c3#diff-56233c94d6ef91a0544ba6661635b959d586b2af8f1440e39df705bde11d77baL173

without testing or E2E
So ok for me to revert it back

keshavbhatt (approved):
The bump happened in tag 4.1.1 https://github.com/zextras/carbonio-mailbox/blob/4.1.1/pom.xml

before this it was 2.5.2

PR #300, opened 2023-08-02
feat: [CO-784] add documentation URL attributes

inno-dima:
**What has changed:**

- Added LDAP attributes in attrs.xml and rights-domainadmin.xml
- Added JSON configuration for LDAP update script

### DIFF
--- a/store/ldap/src/updates/attrs/1690904855.json
+++ b/store/ldap/src/updates/attrs/1690904855.json
@@ -1,0 +1,6 @@
+{
+  "zimbra_globalconfig": [
+    "carbonioAdminDocumentationUrl",
+    "carbonioUserDocumentationUrl"
+  ]
+}
--- a/common/src/main/java/com/zimbra/common/account/ZAttrProvisioning.java
+++ b/common/src/main/java/com/zimbra/common/account/ZAttrProvisioning.java
@@ -2735,6 +2735,14 @@ public class ZAttrProvisioning {
     public static final String A_c = "c";
 
     /**
+     * Link to the Carbonio Admin Documentation
+     *
+     * @since ZCS 23.9.0
+     */
+    @ZAttr(id=3135)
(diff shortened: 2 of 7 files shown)


frisonisland (approved):

store/ldap/src/updates/attrs/1690904855.json
```diff
@@ -0,0 +1,6 @@
+{
+  "zimbra_globalconfig": [
```
keshavbhatt, line 2: We don't need migration, since these are new attributes.

keshavbhatt (changes requested):
Reuqested a change everything else look good.

store/ldap/src/updates/attrs/1690904855.json
frisonisland: Are you sure? The documentation suggest the opposite. I thought the update was meant to add a default value to new attributes created on existing entities
See: https://github.com/zextras/carbonio-mailbox/blob/devel/store/ldap/src/updates/attrs/ReadMe.md

Polpetta (approved):
Everything is fine after the migration thing with LDAP is settled

store/ldap/src/updates/attrs/1690904855.json
keshavbhatt: Yes its true. But as discussed these attributes will probably not have any default value.

frisonisland: It was confirmed we have a default value, but we have to wait for it to be provided!

keshavbhatt (changes requested):
You need to build the project after updating attrs.xml and commit whatever changed, usually the build will generate and update some files. They need to be commited.

Other than this, I asked for minor cleanup of ReadMe file which was updated.

keshavbhatt (approved):

aheeva-yuliya (approved):

PR #301, opened 2023-08-03
feat: [CO-772] Allow admins to specify custom domain login and logout URL

keshavbhatt:
### Whats has changed:

**Added new attributes to manage custom domain login and logout URL**
    - carbonioWebUILoginURL
    - carbonioWebUILogoutURL
    - carbonioAdminUILoginURL
    - carbonioAdminUILogoutURL 
    
**All above mentioned attributes are:**

- optional in `globalConfig` and `domain` entries
- returned as part of the `GetDomainInfo` SOAP call
- are modifiable by a `domain admin` and `global admin` ofcourse

**New proxyconfgen variables to** 

support redirection of web login and logout for both default and virtualhost templates all defaulting to `/static/login/` path.
- for handling login `web.carbonio.[webui|admin].login.url.[default|vhost]`
- for handling logout `web.carbonio.[webui|admin].logout.redirect.[default|vhost]`

- Urls stored in these attributes are aslo added to connect-src directive of CSP header to allow javascript based redirection and fetching.

**Related PRs:**
- https://github.com/zextras/carbonio-proxy/pull/79

**Note:** the redirection of logout is handled by othe carbonio components like UI and Auth.

**Testing:**
- Unit tests were added to verify the behavior of newly added variables.
- Manual testing.


store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
```diff
@@ -0,0 +1,155 @@
...
+            + " https://auth.test.com https://auth.test.com/logout; script-src 'self';;";
+    assertEquals(expectedCspValue, addHeadersVar.confValue());
+  }
+
+  @Test
+  void parseHeaderLine_should_parseHeaderCorrectly() {
+    String headerLine1 = "Content-Security-Policy: default-src 'self';";
+    String headerLine2 = "X-Frame-Options: DENY";
+
+    AddHeadersVar addHeadersVar = new AddHeadersVar(null, null, null, null);
```
frisonisland, line 61: Why all these null? It means the class doesn't need the dependencies probably

store/src/test/java/com/zimbra/cs/util/proxyconfgen/WebCustomLogoutRedirectVarTest.java
```diff
@@ -0,0 +1,50 @@
...
+    String attribute = ZAttrProvisioning.A_carbonioAdminUILogoutURL;
+    Object defaultValue = "/static/login/";
+    String description = "Custom logout redirect URL for web proxy";
+    String customUrl = "https://custom-logout-url.com";
+
+    WebCustomLogoutRedirectVar var =
+        new WebCustomLogoutRedirectVar(keyword, attribute, defaultValue, description, customUrl);
+
+    assertEquals("return 200", var.mValue);
+  }
```
frisonisland, line 49: I see these tests test:
- url empty -> default value
- url not empty -> set actual value

This is not necessary, I would set whatever is passed. 
This would make it non-ambiguous.
Consider choosing the value to set when instantiating the class, passing the correct value.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,155 @@
...
+    }
+
+    // Prepare the new connectSrcDirective as needed
+    final String newConnectSrcDirective = newConnectSrcDirectiveBuilder.toString();
+    if (!connectSrcDirective.equalsIgnoreCase(newConnectSrcDirective)) {
+      return cspValue.replace(connectSrcDirective, newConnectSrcDirective);
+    }
+
+    return "";
+  }
```
frisonisland, line 106: This method name starts with "modifyCsp" but doesn't modify anything, rather it returns a value. So it is a get or the method doesn't perform what it announces.
Also the logout URL conditions are not clear. Probably a smell this method is doing too much.

frisonisland (changes requested):
Code can be cleaned up, I think it would be nice to check together and make it more maintainable

store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
```diff
@@ -0,0 +1,155 @@
...
+    MailboxTestUtil.initServer();
+  }
+
+  @AfterEach
+  public void tearDown() throws Exception {
+    MailboxTestUtil.clearData();
+  }
+
+  @Test
+  void update_should_modifyCspHeaderValueCorrectly() throws ServiceException, ProxyConfException {
```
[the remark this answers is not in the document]
keshavbhatt: I am following `should_expectedBehavior_whenCondition` unit test naming convention as it is the most accepted and practical one. The one that we were using was hard to read and hard to write, We can have a quick talk about this and think about adopting it everywhere. Anyway, there are many conventions there to name your test methods but the one we were using is documented no where. I persoanlly had reading and writing issues with that, and if you will see adding those underscore as separator fixes the issue making it easily readable.

store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
keshavbhatt: Those values are not required for this test.

store/src/test/java/com/zimbra/cs/util/proxyconfgen/WebCustomLogoutRedirectVarTest.java
keshavbhatt: To change this behaviour we need to change the logic of ProxyConfVar which this class extends.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
keshavbhatt: Updated the name to  `generateModifiedCspHeaderValue`, extracted cspUpdate logic to `generateModifiedConnectSrcDirectiveWithCustomUrls`.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
```diff
@@ -11,7 +11,7 @@ class DomainAttrExceptionItem extends DomainAttrItem {
   ProxyConfException exception;
 
   public DomainAttrExceptionItem(ProxyConfException e) {
-    super(null, null, null, null, null, null, null, null, null);
+    super(null, null, null, null, null, null, null, null, null, null, null, null, null);
```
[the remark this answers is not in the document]
keshavbhatt: DomainAttrItem now uses builder, this custom exception was updated accordingly.

store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
```diff
@@ -0,0 +1,155 @@
...
+    assertTrue(AddHeadersVar.isValidSrcDirectiveUrl("https://www.test.com:99999"));
+  }
+
+  @Test
+  void isValidSrcDirectiveUrl_should_returnFalseForInvalidUrls() {
+    assertFalse(AddHeadersVar.isValidSrcDirectiveUrl("invalid_url"));
+    assertFalse(AddHeadersVar.isValidSrcDirectiveUrl("http:/test.com"));
+    assertFalse(AddHeadersVar.isValidSrcDirectiveUrl("https:/test.com"));
+    assertFalse(AddHeadersVar.isValidSrcDirectiveUrl("https://www.test.com,https://test.com"));
+    assertFalse(AddHeadersVar.isValidSrcDirectiveUrl("https://www.test.com https://test.com"));
```
frisonisland, line 151: Use a parameterized test, you can pass urls as arguments.
Also I see we named the test as "<methodName>_should_<condition+expectation>".
We could name it as "shouldReturnFalse_WhenInvalidUrl".

store/src/test/java/com/zimbra/cs/util/proxyconfgen/WebCustomLogoutRedirectVarTest.java
```diff
@@ -0,0 +1,50 @@
...
+    final String keyword = "web.custom.logout.redirect";
+    final String attribute = ZAttrProvisioning.A_carbonioAdminUILogoutURL;
+    final Object defaultValue = "/static/login/";
+    final String description = "Custom logout redirect URL for web proxy";
+    final String customUrl = "";
+
+    final WebCustomLogoutRedirectVar var =
+        new WebCustomLogoutRedirectVar(keyword, attribute, defaultValue, description, customUrl);
+
+    assertEquals("return 307 /static/login/", var.mValue);
```
frisonisland, line 34: This test tells me we are setting a default value on constructor if value empty.
Generally is better to just assign things in Constructor, with no logic.
We can achieve the same by making the constructor private and using a factory method that checks the emptiness.

store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
keshavbhatt: This is exactly the same naming convention that is used across other method names.

methodUnderTest_should_expectedBehavior_whenCondition.

keshavbhatt: Done,
But, I dont know why we want to take this freedom away :D

store/src/test/java/com/zimbra/cs/util/proxyconfgen/WebCustomLogoutRedirectVarTest.java
keshavbhatt: removed update method call from constructor. leaving with no logic check

store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
frisonisland: It's the same concept as mixing snake case and camel case in the end, team decision in the end + naming consistency.
It would be odd if in a next PR someone decides to use another naming convention for example.
If we care about the name being displayed during tests consider using: https://junit.org/junit5/docs/current/api/org.junit.jupiter.api/org/junit/jupiter/api/DisplayName.html
since we adopted junit jupiter

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,164 @@
...
+   * Content-Security-Policy.
+   *
+   * @param url The URL to be validated.
+   * @return {@code true} if the URL is a valid source directive URL, otherwise {@code false}.
+   * @author Keshav Bhatt
+   * @since 23.9.0
+   */
+  static boolean isValidSrcDirectiveUrl(String url) {
+    return Pattern.matches(
+        "^(https?://(w{3}\\.)?)?((\\*\\.)?\\w+(-\\w+)*\\.\\w+(\\.[a-zA-Z]+)*(:\\d{1,5})?(/\\*|/\\w*)*(\\??(.+=.*)?(&.+=.*)?)?)?$",
```
frisonisland, line 37: I think for all utilities method in ProxyConfGen context it could be a good idea to move them to a ProxyConfGenUtil class. This way we will slowly slim down ProxyConfGen and related class.
Also the name of this class is bit weird IMHO, why AddHeadersVar and not HeadersVar? I know we didn't create it, but does it represent a Header or an Add Header operation?

keshavbhatt: It is named like this maybe because it is supposed to generate "add_header is a nginx directive". I agree to move this method to ProxyConfGenUtil.

keshavbhatt: done.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,164 @@
...
+      // Handle CSP header
+      if (header.key.equalsIgnoreCase("Content-Security-Policy")) {
+        String newCspValue = generateModifiedCspHeaderValue(header.value);
+        if (!newCspValue.isEmpty()) {
+          header = new KeyValue(header.key, newCspValue);
+        }
       }
-      directives.add(headers[i]);
-      i++;
+      directives.add(header);
```
frisonisland: I will ask to try to move this logic outside, it is weird having a class update itself and leads to side effects. I prefer returning a value from a method rather then void and update its internal value. This because it is unclear you have to call the update() to get some value later, it is more natural to call a get() (which can call an update status internally ofc).
Also I have no idea how these "mValues" are being used and. I think there a re lots of options to make the operation more clear, these are just raw ideas:
1) Creating a HeaderVar interface with a get() method that returns the requested header as string for nginx. This is basically a
https://docs.oracle.com/javase/8/docs/api/java/util/function/Supplier.html
We can add special parsing logic for special headers or just return the identity (output equal to input).
2) Rename this class as ProxyHeaders with a get() method. When calling the get() it returns parsed headers with correct values (the mValues).
This ismore or less the current update() except it returns values.
Since the topic is wide let's discuss together IRL

keshavbhatt: Agree with points but also notice that this class inherits ProxyConfVar, which has its own trait and changing them is out of scope if this task. We can approach refactoring this in near future.

If the behavior is not clear we can talk about it for sure.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
```diff
@@ -1,17 +1,60 @@
...
+   * Get the associated exception.
+   *
+   * @return The associated {@link ProxyConfException}.
+   */
+  public ProxyConfException getException() {
+    return exception;
+  }
+
+  /** Builder class for constructing {@link DomainAttrExceptionItem} instances. */
+  public static class Builder extends DomainAttrItem.Builder {
```
Polpetta, line 37: I have several questions about this Builder class.
1. Does it need to be public? With IntelliJ I don't see any usage. If the design of this builder (and relative class) is to be used inside its package only, then the builder too should be marked as `package-private` (i.e remove the `public` attribute)
2. Add `@author` tag on the builder


On a final note, I see you're creating a hierarchy of Builders and of DomainAttrItem. I've checked it out and I think this is a slippery slope for something that will be difficult to maintain in the future. Instead of extending DomainAttrItem I'd make use of `Try` object, encapsulating the exception if something bad happens, and returning the expected objet if everything is fine. This way, we avoid also the `instanceof` that are present in `ProxyConfGen` that violates SOLID principles.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
```diff
@@ -1,17 +1,60 @@
...
-  public DomainAttrExceptionItem(ProxyConfException e) {
-    super(null, null, null, null, null, null, null, null, null);
-    this.exception = e;
+    /**
+     * Build a {@link DomainAttrExceptionItem} instance with an associated exception.
+     *
+     * @param exception The exception to associate with the item.
+     * @return A new {@link DomainAttrExceptionItem} instance.
+     */
+    public DomainAttrExceptionItem buildWithException(ProxyConfException exception) {
```
Polpetta, line 56: This is super fragile. Another user will end up confused about calling `build` or `buildWithException`. I suggest overriding `build` and directly return `DomainAttrExceptionItem` instead.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
```diff
@@ -1,17 +1,60 @@
...
+ * Represents a domain attribute exception item, which is a specialized form of {@link
+ * DomainAttrItem} that includes an associated exception.
  *
- * @author jiankuan
+ * <p>The visit of {@link com.zimbra.cs.account.ldap.LdapProvisioning} can't throw the exception
+ * out. Therefore, uses this special item to indicate exception.
+ *
+ * @author jiankuan, Keshav Bhatt
  */
 class DomainAttrExceptionItem extends DomainAttrItem {
```
Polpetta, line 12: This hierarchy makes little sense imo. An Exception shouldn't extend a parent class. This is just exploiting the Java type system, but it is not maintainable nor a sane design choice.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
```diff
@@ -8,34 +8,170 @@
...
+  private final String clientCertMode;
+  private final String clientCertCa;
+  private final String[] rspHeaders;
+  private final String cspHeader;
+  private final String webUiLoginUrl;
+  private final String webUiLogoutUrl;
+  private final String adminUiLoginUrl;
+  private final String adminUiLogoutUrl;
+
+  DomainAttrItem(Builder builder) {
```
Polpetta, line 25: You shouldn't pass the builder here. It was correct as it was done previously. Please revert this and pass all the arguments to the constructor. The builder itself will take care to call the constructor with all the correct parameters (also fix the constructor visibility to `private`)

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
```diff
@@ -8,34 +8,170 @@
...
+    private String clientCertMode;
+    private String clientCertCa;
+    private String[] rspHeaders;
+    private String cspHeader;
+    private String webUiLoginUrl;
+    private String webUiLogoutUrl;
+    private String adminUiLoginUrl;
+    private String adminUiLogoutUrl;
+
+    public Builder(String domainName, String virtualHostname, String virtualIPAddress) {
```
Polpetta, line 117: In order to set mandatory parameters instead of using the builder constructor consider instead the Step Builder pattern: https://www.javacodegeeks.com/2013/05/building-smart-builders.html

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
```diff
@@ -8,34 +8,170 @@
...
+
+  public String getAdminUiLoginUrl() {
+    return adminUiLoginUrl;
+  }
+
+  public String getAdminUiLogoutUrl() {
+    return adminUiLogoutUrl;
+  }
+
+  public static class Builder {
```
Polpetta, line 101: `DomainAttrItem` is not public but the builder is. Is this intended behavior? If not fix the visibility please (all the methods inside the class too)

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -56,6 +81,7 @@ public class ProxyConfGen {
   static final String ZIMBRA_SSL_UPSTREAM_ZX_NAME = "zx_ssl";
   static final int ZIMBRA_UPSTREAM_ZX_PORT = 8742;
   static final int ZIMBRA_UPSTREAM_SSL_ZX_PORT = 8743;
+  static final String DEFAULT_WEB_LOGIN_PATH = "/static/login/";
```
Polpetta, line 84: I don't know why this variable is package private. It can be put as `private` instead.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -79,27 +108,27 @@ public class ProxyConfGen {
...
   private static boolean mGenConfPerVhn = false;
   private static boolean hasCustomTemplateLocationArg = false;
-  private static final String CERTBOT = mWorkingDir + "/libexec/certbot";
-  private static final String CERTBOT_WORKING_DIR =
-      mWorkingDir + "/common/certbot/etc/letsencrypt/live/";
-  private static final String CERT = "/fullchain.pem";
-  private static final String KEY = "/privkey.pem";
 
   static {
+    mConfIncludesDir = mConfDir + File.separator + mIncDir;
```
Polpetta, line 131: I don't know why you moved this line here. Please revert it as it was, that was the correct way. Doing stuff in static block is strongly discouraged, and I'd seek ways to eliminate this block instead of making it bigger.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -251,6 +284,10 @@ private static List<DomainAttrItem> loadDomainReverseProxyAttrs() throws Service
               entry.getMultiAttr(ZAttrProvisioning.A_zimbraReverseProxyResponseHeaders);
           String cspRspHeader =
               entry.getAttr(ZAttrProvisioning.A_carbonioReverseProxyResponseCSPHeader, "");
+          String webUiLoginUrl = entry.getAttr(ZAttrProvisioning.A_carbonioWebUILoginURL, "");
```
Polpetta, line 287: Mark it as `final`

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -251,6 +284,10 @@ private static List<DomainAttrItem> loadDomainReverseProxyAttrs() throws Service
...
+          String webUiLogoutUrl = entry.getAttr(ZAttrProvisioning.A_carbonioWebUILogoutURL, "");
```
Polpetta, line 288: Mark it as `final`

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -251,6 +284,10 @@ private static List<DomainAttrItem> loadDomainReverseProxyAttrs() throws Service
...
+          String adminUiLoginUrl = entry.getAttr(ZAttrProvisioning.A_carbonioAdminUILoginURL, "");
```
Polpetta, line 289: Mark it as `final`

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -251,6 +284,10 @@ private static List<DomainAttrItem> loadDomainReverseProxyAttrs() throws Service
...
+          String adminUiLogoutUrl = entry.getAttr(ZAttrProvisioning.A_carbonioAdminUILogoutURL, "");
```
Polpetta, line 290: Mark it as `final`

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -663,43 +702,83 @@ private static void fillVarsWithDomainAttrs(DomainAttrItem item) throws ProxyCon
...
-      responseHeadersList.add(item.rspHeaders[i]);
+    for (i = 0; i < item.getRspHeaders().length; i++) {
+      responseHeadersList.add(item.getRspHeaders()[i]);
     }
-    if (!item.cspHeader.isBlank()) {
-      responseHeadersList.add(item.cspHeader);
+    if (!item.getCspHeader().isBlank()) {
+      responseHeadersList.add(item.getCspHeader());
     }
+    // Add domain the response headers list variables for web
```
Polpetta, line 711: Please remove code comments. They make the code less maintainable.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
```diff
@@ -663,43 +702,83 @@ private static void fillVarsWithDomainAttrs(DomainAttrItem item) throws ProxyCon
...
-            "add_header directive for vhost web proxy"));
-
-    LOG.debug("Updating Default Domain Variable Map");
+            "add_header directive for vhost web proxy",
+            Map.of(
+                ZAttrProvisioning.A_carbonioWebUILoginURL, item.getWebUiLoginUrl(),
+                ZAttrProvisioning.A_carbonioWebUILogoutURL, item.getWebUiLogoutUrl(),
+                ZAttrProvisioning.A_carbonioAdminUILoginURL, item.getAdminUiLoginUrl(),
+                ZAttrProvisioning.A_carbonioAdminUILogoutURL, item.getAdminUiLogoutUrl())));
+    // Add domain custom login and logout variables for webUi and AdminUi
```
Polpetta, line 723: Please remove code comments. They make the code less maintainable.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/WebCustomLoginUrlVar.java
```diff
@@ -0,0 +1,33 @@
...
+ *
+ * @author Keshav Bhatt
+ * @see ProxyConfVar
+ * @since 23.9.0
+ */
+class WebCustomLoginUrlVar extends ProxyConfVar {
+
+  private final String mCustomUrl;
+
+  public WebCustomLoginUrlVar(
```
Polpetta, line 16: This constructor shouldn't be `public`, instead `package-private`. Please fix the visibility accordingly.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/WebCustomLogoutRedirectVar.java
```diff
@@ -0,0 +1,43 @@
...
+ *
+ * @author Keshav Bhatt
+ * @see ProxyConfVar
+ * @since 23.9.0
+ */
+class WebCustomLogoutRedirectVar extends ProxyConfVar {
+
+  private final String mCustomUrl;
+
+  public WebCustomLogoutRedirectVar(
```
Polpetta, line 24: Fix visibility to `package-private`.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,132 @@
...
-      Matcher matcher = RE_HEADER.matcher(hdr);
-      if (matcher.matches()) {
-        headers[i] = new KeyValue(matcher.group(1), matcher.group(2));
-      } else {
-        headers[i] = new KeyValue(hdr);
+
+    for (String headerLine : responseHeaders) {
+      KeyValue header = ProxyConfUtil.parseHeaderLine(headerLine);
+
+      // Handle CSP header
```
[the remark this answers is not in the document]
keshavbhatt: done

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,132 @@
...
+      if (header.key.equalsIgnoreCase("Content-Security-Policy")) {
+        String newCspValue = generateModifiedCspHeaderValue(header.value);
```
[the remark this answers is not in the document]
keshavbhatt: done

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,132 @@
...
+   * connect-src directive if they don't exist. If no connect-src directive is found, an empty
+   * string is returned.
+   *
+   * @param cspValue The original Content-Security-Policy header value to modify.
+   * @return The modified Content-Security-Policy header value with added custom URLs to the
+   *     connect-src directive, or an empty string if the connect-src directive is missing.
+   * @author Keshav Bhatt
+   * @since 23.9.0
+   */
+  String generateModifiedCspHeaderValue(String cspValue) {
```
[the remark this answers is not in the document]
keshavbhatt: done

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,132 @@
...
+   * Extracts the connect-src directive from the Content-Security-Policy header. The method uses a
+   * case-insensitive regular expression pattern to match the directive.
+   *
+   * @param cspHeader The Content-Security-Policy header to extract the connect-src directive from.
+   * @return The connect-src directive of the Content-Security-Policy header if found, or an empty
+   *     string if not found.
+   * @author Keshav Bhatt
+   * @since 23.9.0
+   */
+  String extractConnectSrcDirective(String cspHeader) {
```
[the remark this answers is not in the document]
keshavbhatt: done, moved to `ProxyConfUtil`

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
```diff
@@ -1,17 +1,60 @@
...
- * The visit of LdapProvisioning can't throw the exception out. Therefore, uses this special item to
- * indicate exception.
+ * Represents a domain attribute exception item, which is a specialized form of {@link
+ * DomainAttrItem} that includes an associated exception.
  *
- * @author jiankuan
+ * <p>The visit of {@link com.zimbra.cs.account.ldap.LdapProvisioning} can't throw the exception
+ * out. Therefore, uses this special item to indicate exception.
+ *
+ * @author jiankuan, Keshav Bhatt
```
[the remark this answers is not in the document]
keshavbhatt: done.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
```diff
@@ -1,17 +1,60 @@
...
+   * @param builder The builder used to construct the parent DomainAttrItem instance.
+   * @param exception The exception associated with this DomainAttrExceptionItem.
+   */
+  private DomainAttrExceptionItem(DomainAttrItem.Builder builder, ProxyConfException exception) {
+    super(builder);
+    this.exception = exception;
+  }
+
+  /**
+   * Get the associated exception.
```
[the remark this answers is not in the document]
keshavbhatt: done

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
```diff
@@ -1,17 +1,60 @@
...
+   *
+   * @return The associated {@link ProxyConfException}.
+   */
+  public ProxyConfException getException() {
```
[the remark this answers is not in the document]
keshavbhatt: done.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
keshavbhatt: removed the builder class, due to no usages.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
keshavbhatt: removed, as not useful anymore.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
keshavbhatt: You are right, the reason is explained in the javadoc :)

`The visit of com.zimbra.cs.account.ldap.LdapProvisioning can't throw the exception out. Therefore, uses this special item to indicate exception.`

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
keshavbhatt: fixed.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
keshavbhatt: fixed.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/ProxyConfGen.java
keshavbhatt: This was done to avoid illegal forward referencing of variable that composes mConfIncludesDir. This issue was happening due to inteiij rearraging the code. This will have no major impact.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
keshavbhatt: using fluent complete builder.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,132 @@
...
+public class AddHeadersVar extends ProxyConfVar {
 
-  private final ArrayList<String> rhdr;
-  private int i;
+  private final ArrayList<String> responseHeaders;
+  private final Map<String, String> customLoginLogoutUrls;
 
-  public AddHeadersVar(String key, ArrayList<String> rhdr, String description) {
+  public AddHeadersVar(
+      String key,
```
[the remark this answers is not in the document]
frisonisland: @keshavbhatt  we discussed AddHeadersVar is the add_header directive of Nginx. We can put it in JavaDoc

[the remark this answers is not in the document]
frisonisland: Also "key" in this case is header name, right? I would just make it clear with variable naming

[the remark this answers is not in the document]
keshavbhatt: Provided javadoc. It is called keyword now and is aligned with the name suggested by the base class. 

"keyword is word for this variable that will be used to explode the value in nginx templates"

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
```diff
@@ -2,50 +2,132 @@
...
+      ArrayList<String> responseHeaders,
+      String description,
```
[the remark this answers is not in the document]
keshavbhatt: > Why is description needed
This represents the description of this ProxyConfVar.

> Also: description of what?
description of this ProxyConfVar offcourse.

Also note that these things are not made up or added for fun. Reading and understanding the code can help understand why these things are involved.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVar.java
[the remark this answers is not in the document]
keshavbhatt: CSPHeaderVar is not something that represent a proxyconfvar. It is something that is a header which is already covered by add_header proxyconfvar. hence it is handled within the add_header.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrExceptionItem.java
```diff
@@ -1,17 +1,60 @@
...
+  /**
+   * Constructs a new {@link DomainAttrExceptionItem} instance.
+   *
+   * @param builder The builder used to construct the parent DomainAttrItem instance.
+   * @param exception The exception associated with this DomainAttrExceptionItem.
+   */
+  private DomainAttrExceptionItem(DomainAttrItem.Builder builder, ProxyConfException exception) {
+    super(builder);
+    this.exception = exception;
+  }
```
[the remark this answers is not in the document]
keshavbhatt: I did not created this exception class. Renaming it is beyond the scope of this task.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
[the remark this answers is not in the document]
keshavbhatt: > Consider using constructor without passing a builder, instead passing each field.

IDK if this already became outdated. If not please let me know what you are refering to. 

Also i checked possibility of making the constructor private, but it causes issues in DomainAttrExceptionItem since it extends DomainAttrItem.

store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
```diff
@@ -0,0 +1,84 @@
...
+    MailboxTestUtil.initServer();
+  }
+
+  @AfterEach
+  public void tearDown() throws Exception {
+    MailboxTestUtil.clearData();
+  }
+
+  @Test
+  void update_should_modifyCspHeaderValueCorrectly() throws ServiceException, ProxyConfException {
```
frisonisland, line 28: GIven these two tests are nearly the same, is it possible to make them more compact? Perhaps using parameterized tests we can pass one or more urls in input

store/src/test/java/com/zimbra/cs/util/proxyconfgen/WebCustomLogoutRedirectVarTest.java
```diff
@@ -0,0 +1,50 @@
...
+import com.zimbra.cs.mailbox.MailboxTestUtil;
+import org.junit.jupiter.api.AfterEach;
+import org.junit.jupiter.api.BeforeAll;
+import org.junit.jupiter.api.Test;
+
+class WebCustomLogoutRedirectVarTest {
+
+  @BeforeAll
+  public static void init() throws Exception {
+    MailboxTestUtil.initServer();
```
frisonisland, line 15: Why do we need to spin up the whole Mailbox? Perhaps these classes are using the Provisioning in a static way underneath? If so, is it possible to provide the Provisioning in constructor so we are able to Unit Test the class?
Asking because I see we are not doing any setup operation on variables.
I think a test where we would need to start the mailbox should be for example: write web.custom.logout.redirect variable to LDAP, check the template file from proxyconfgen contains the correct value

frisonisland (changes requested):
I have added some comments, understanding that for now the ProxyConfGen design will be still the same, as it will impact too many classes

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
```diff
@@ -1,41 +1,233 @@
...
+
+    VirtualIPAddressStep withVirtualHostname(String virtualHostname);
+  }
+
+  interface VirtualIPAddressStep {
+
+    FinalStep withVirtualIPAddress(String virtualIPAddress);
+  }
+
+  interface FinalStep {
```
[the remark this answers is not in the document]
keshavbhatt: @frisonisland we are implementing Step builder. 
https://www.svlada.com/step-builder-pattern/
https://medium.com/@castigliego/step-builder-pattern-3bcac4eaf9e8

store/conf/attrs/attrs.xml
```diff
@@ -10122,4 +10122,20 @@ TODO: delete them permanently from here
...
+
+<attr id="3136" name="carbonioWebUILoginURL" type="string" max="256" cardinality="single" optionalIn="globalConfig,domain" flags="domainInfo,domainAdminModifiable" requiresRestart="nginxproxy" since="23.9.0">
+  <desc>Login URL for Carbonio web client to send the user to upon failed login, auth expired, or no/invalid auth</desc>
+</attr>
+
+<attr id="3137" name="carbonioWebUILogoutURL" type="string" max="256" cardinality="single" optionalIn="globalConfig,domain" flags="domainInfo,domainAdminModifiable" requiresRestart="nginxproxy" since="23.9.0">
+  <desc>Logout URL for Carbonio web client to send the user to upon explicit logging out</desc>
+</attr>
+
+<attr id="3138" name="carbonioAdminUILoginURL" type="string" max="256" cardinality="single" optionalIn="globalConfig,domain" flags="domainInfo,domainAdminModifiable" requiresRestart="nginxproxy" since="23.9.0">
```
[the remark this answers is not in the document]
keshavbhatt: Yes.

store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
keshavbhatt: > GIven these two tests are nearly the same, is it possible to make them more compact? Perhaps using parameterized tests we can pass one or more urls in input

It will look something like the following, which is more confusing then keeping them separeate i think.

```
class AddHeadersVarTest {
  private Provisioning provisioning;

  @BeforeEach
  void setUp() {
    provisioning = mock(Provisioning.class);
  }

  @ParameterizedTest
  @MethodSource("testData")
  void update_should_modifyCspHeaderValueCorrectly(
      String key, ArrayList<String> responseHeaders, String description,
      Map<String, String> customLogInLogoutUrlValueMap, String expectedCspValue)
      throws ServiceException, ProxyConfException {

    final AddHeadersVar addHeadersVar =
        new AddHeadersVar(provisioning, key, responseHeaders, description,
            customLogInLogoutUrlValueMap);

    addHeadersVar.update();

    assertEquals(expectedCspValue, addHeadersVar.confValue());
  }

  private static Stream<Object[]> testData() {
    return Stream.of(
        new Object[]{
            "web.add.headers.vhost",
            createResponseHeadersList(),
            "add_header directive for vhost web proxy",
            createCustomLogInLogoutUrlValueMap(),
            "add_header Content-Security-Policy default-src 'self'; connect-src 'self' *.test.tools"
                + " https://auth.test.com https://auth.test.com/logout; script-src 'self';;"
        },
        new Object[]{
            "web.add.headers.default",
            createResponseHeadersList(),
            "add_header directive for vhost web proxy",
            createCustomLogInLogoutUrlValueMapWithAdminUrls(),
            "add_header Content-Security-Policy default-src 'self'; connect-src 'self'"
                + " *.test.tools https://auth.test.com https://admin-auth.test.com/logout"
                + " https://auth.test.com/logout https://admin-auth.test.com; script-src 'self';;"
        }
    );
  }

  private static ArrayList<String> createResponseHeadersList() {
    final ArrayList<String> responseHeaders = new ArrayList<>();
    responseHeaders.add(
        "Content-Security-Policy: default-src 'self'; connect-src 'self' *.test.tools;"
            + " script-src 'self';");
    return responseHeaders;
  }

  private static Map<String, String> createCustomLogInLogoutUrlValueMap() {
    final Map<String, String> customLogInLogoutUrlValueMap = new HashMap<>();
    customLogInLogoutUrlValueMap.put(
        ZAttrProvisioning.A_carbonioWebUILoginURL, "https://auth.test.com");
    customLogInLogoutUrlValueMap.put(
        ZAttrProvisioning.A_carbonioWebUILogoutURL, "https://auth.test.com/logout");
    return customLogInLogoutUrlValueMap;
  }

  private static Map<String, String> createCustomLogInLogoutUrlValueMapWithAdminUrls() {
    final Map<String, String> customLogInLogoutUrlValueMap = new HashMap<>();
    customLogInLogoutUrlValueMap.put(
        ZAttrProvisioning.A_carbonioWebUILoginURL, "https://auth.test.com");
    customLogInLogoutUrlValueMap.put(
        ZAttrProvisioning.A_carbonioWebUILogoutURL, "https://auth.test.com/logout");
    customLogInLogoutUrlValueMap.put(
        ZAttrProvisioning.A_carbonioAdminUILoginURL, "https://admin-auth.test.com");
    customLogInLogoutUrlValueMap.put(
        ZAttrProvisioning.A_carbonioAdminUILogoutURL, "https://admin-auth.test.com/logout");
    return customLogInLogoutUrlValueMap;
  }
}
```

store/src/test/java/com/zimbra/cs/util/proxyconfgen/WebCustomLogoutRedirectVarTest.java
keshavbhatt: resolved in IRL.

store/src/main/java/com/zimbra/cs/util/proxyconfgen/DomainAttrItem.java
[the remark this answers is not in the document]
frisonisland: Thx, talked IRL and now makes more sense to me

store/src/test/java/com/zimbra/cs/util/proxyconfgen/AddHeadersVarTest.java
keshavbhatt: You mean swaping the key? If yes,
the key is only to store a value, value can be anything. We use appropriate key in our templates to expload whatevery is stored in it by the proxyconfgen.

frisonisland: Also be careful about map ordering. The example provided above fails due to computation of values in the implementation: sometimes it returns https://auth.test.com before https://auth.test.com/logout and viceversa.
Can we be sure the order of values is always the same?

keshavbhatt: I am fine with use of parameterized tests when the test data is almost same, but in this case they are not just urls but different types of URLs. Offcourse this setup will work but the issue i see with this is: It doesn't make clear distinction between the test data.  I am also considering future of this test class where will may add more tests.
I decide to keep them separate.

frisonisland: Another thing: these tests are telling us a lot about how we use our classes. 
AddHeadersVar uses customLoginLogoutUrls to iterate over values(), which is an array, so why not directly pass customLoginLogoutUrls as array instead of map? -> Key are never accessed + this way order of values is preserved

keshavbhatt: Probably the map was introduces in the past when the implementaiton was different. The featuer reuqest changed later and the map remained there :D

Anyway, the latest changes are using List.

frisonisland (approved):

Polpetta (approved):

PR #302, opened 2023-08-04
feat: [CO-779] Use latest Preview SDK

inno-dima:
**What has changed**

- bump preview-sdk version to 1.0.2
- Added GIF support in PreviewServlet.java

### DIFF
--- a/pom.xml
+++ b/pom.xml
@@ -458,7 +458,7 @@
       <dependency>
         <groupId>com.zextras.carbonio.preview</groupId>
         <artifactId>carbonio-preview-sdk</artifactId>
-        <version>1.0.1</version>
+        <version>1.0.2</version>
       </dependency>
       <dependency>
         <groupId>org.mockito</groupId>
--- a/store/src/main/java/com/zimbra/cs/service/PreviewServlet.java
+++ b/store/src/main/java/com/zimbra/cs/service/PreviewServlet.java
@@ -58,7 +58,7 @@ import org.eclipse.jetty.http.HttpStatus;
  *
  * <pre>
  *
- *   Based on Carbonio Preview SDK 0.2.4
+ *   Based on Carbonio Preview SDK 1.0.2
  *
  *   The API is the almost same as of preview service(https://zextras.atlassian.net/wiki/spaces/SW/pages/2353430753/Preview+API)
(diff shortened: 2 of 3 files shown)


store/src/main/java/com/zimbra/cs/service/PreviewServlet.java
```diff
@@ -728,7 +732,9 @@ enum Format {
     JPEG,
 
     @JsonProperty("png")
-    PNG
+    PNG,
+    @JsonProperty("gif")
+    GIF
```
frisonisland, line 737: Tests for this feature?

inno-dima: I'll do it, but I'm not sure how to test it properly. Let's discuss it on Monday

frisonisland (changes requested):
It is not clear what is needed to add GIF support. 
Should we add tests for parsing?
Consider starting from some tests that define the use case. Also I see this class is way too big, definitely a smell, consider refactoring it if it helps testing! (I'm sure it will).
For example: PreviewQueryParameters class that is used in the client call. It might help splitting the servlet in Controller and another Service/Logic classs

pom.xml
```diff
@@ -458,7 +458,7 @@
       <dependency>
         <groupId>com.zextras.carbonio.preview</groupId>
         <artifactId>carbonio-preview-sdk</artifactId>
-        <version>1.0.1</version>
+        <version>1.0.2</version>
```
Polpetta, line 461: Please move the version as a POM variable. It makes easy to see all the dependency versions since we'll gradually group them all in one place.

Polpetta (changes requested):
As @frisonisland said, I agree to the fact that there's the necessity to add tests for the new feature.

I also added a minor improvement.

store/src/main/java/com/zimbra/cs/service/PreviewServlet.java
frisonisland: I was thinking that perhaps we can add a test that attests the gif output_format is working correctly

Polpetta (approved):

frisonisland (approved):

keshavbhatt (approved):

PR #307, opened 2023-08-18
feat: [CO-807] update mailboxd_java_options for JDK17

keshavbhatt:
**What has changed:**
- update `mailboxd_java_options` in `LC.java` to allow reflection access at runtime **(this ships update in fresh installs)**
- update `mailboxd_java_options` in `LC.java` adding `java.security.egd` to use `file:/dev/./urandom` as SecureRandom source
- post-upgrade scripts were added to carbonio-common-appserver-conf package to handle the update of `mailboxd_java_options` **(this ships update on existing installs)**

### DIFF
--- a/packages/common-appserver-conf/PKGBUILD
+++ b/packages/common-appserver-conf/PKGBUILD
@@ -202,3 +202,61 @@ package() {
   install -D "store-conf/conf/rights/user-rights.xml" \
     "${pkgdir}/opt/zextras/conf/rights/user-rights.xml"
 }
+
+postinst:apt() {
+  # Update LC mailboxd_java_options for JDK17 on post-upgrade
+  if [ "$1" = "configure" ] && [ ! -z "$2" ]; then
+    echo -n "*Updating mailboxd_java_options for JDK17..."
+    java_options=$(su - zextras -c "/opt/zextras/bin/zmlocalconfig -m nokey mailboxd_java_options 2> /dev/null")
+    new_java_options="$java_options"
+    needs_update=false
+    if ! echo "$java_options" | grep -q -e '-Djava.security.egd'; then
+        new_java_options="$new_java_options -Djava.security.egd=file:/dev/./urandom"
+        needs_update=true
+    fi
+    if ! echo "$java_options" | grep -q -e '--add-opens\ java.base/java.lang=ALL-UNNAMED'; then
+        new_java_options="$new_java_options --add-opens java.base/java.lang=ALL-UNNAMED"
+        needs_update=true
+    fi
+    if $needs_update; then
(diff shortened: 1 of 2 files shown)


frisonisland (approved):

M0Rf30 (approved):

packages/common-appserver-conf/PKGBUILD
```diff
@@ -202,3 +202,61 @@ package() {
...
+    needs_update=false
+    if ! echo "$java_options" | grep -q -e '-Djava.security.egd'; then
+        new_java_options="$new_java_options -Djava.security.egd=file:/dev/./urandom"
+        needs_update=true
+    fi
+    if ! echo "$java_options" | grep -q -e '--add-opens\ java.base/java.lang=ALL-UNNAMED'; then
+        new_java_options="$new_java_options --add-opens java.base/java.lang=ALL-UNNAMED"
+        needs_update=true
+    fi
+    if $needs_update; then
```
Polpetta, line 221: Is this shellcheck approved?

packages/common-appserver-conf/PKGBUILD
```diff
@@ -202,3 +202,61 @@ package() {
...
```
Polpetta, line 250: Same :sweat_smile:

Polpetta (approved):
I've a couple of questions but overall it seems fine to me

packages/common-appserver-conf/PKGBUILD
keshavbhatt: @Polpetta yes, the shell is `sh` in this case and not `bash`

packages/common-appserver-conf/PKGBUILD
keshavbhatt: @Polpetta yes, the shell is sh in this case and not bash

packages/common-appserver-conf/PKGBUILD
Polpetta: Shellcheck can check `sh` stuff too, I'm missing the point of your statement 🤔

keshavbhatt: Yes, I checked with shellcheck, no major issues. The package manager run these scriptlets using sh(Bourne Shell) the (https://www.baeldung.com/linux/sh-vs-bash).

keshavbhatt: Even though most distros uses /bin/sh to point to bash(creating symbolic) or a shell they want to use, i had issues with the bash syntaxted code when i executed it on RH environment. So, i switched to syntax which runs on both bash and pure sh(Bourne Shell).

PR #310, opened 2023-08-24
feat: [CO-778] zimlet code cleanup

keshavbhatt:
**What has changed:**
- removed related SOAP APIs
- deprecated related attributes
- realted LC keys were removed
- removal/update related classes
- locale propeties files cleanup  
- removed stale dependencies

**NOTE for future:**

**We can perform further cleanup, as some part of the code was kept for the following reason.**
- Zimlet Entry was kept : “zimlet”s are LDAP entry types, and many LDAP attributes are present in the system which are expecting this entry in the system. If we remove the zimlet entry type from backend and schema in our coming update, we leave users to the risk where they will have some of ldap entries in locked state(this happens when LDAP schema is not aware of the entry type due to missing definition in schema). For this reason the entry will be kept intact in the codebase.
- Attempting to perform further cleanup affects GetInfo SOAP command, resulting in removal of some zimlet specific fileds(elements,attributes & properties) from the response. This is breaking the Carbonio Web UI and admin UI as they are still looking for these fileds for no obvious reasons. Cleanup in the Rights was also skipped for the same reason.


store/ant-store.xml
```diff
@@ -129,15 +129,15 @@
   <target name="generate-getter">
     <java classname="com.zimbra.cs.account.AttributeManagerUtil" fork="true"
       classpath="${maven_compile_classpath}" failonerror="true">
-      <sysproperty key="carbonio.version" value="${carbonio.buildinfo.all}"/>
```
frisonisland, line 132: I just noticed carbonio.buildinfo.version is never set. and "zimbra" is. What if we declare carbonio.buildinfo.version in place of zimbra.buildinfo.version for the future? I saw it is defined in ant-build-common and never used.

frisonisland (changes requested):
Cleanup looks good to me, just have a suggestion about carbonio version usage in ant

Polpetta (approved):

store/ant-store.xml
keshavbhatt: Yes, What i noticed while working on this PR is that there is no property called carbonio.buildingo.all while there was zimbra.buildinfo.all so for a quick fix i used the it. We need to provide the carbonio one if it is really required.

aheeva-yuliya (approved):

store/ant-store.xml
frisonisland: zimbra.buildinfo.all is not used from what I checked, that's why I suggested to define carbonio.buildinfo.all in place of it

frisonisland (approved):

PR #312, opened 2023-08-28
test: add MailboxSoapTest to test SOAP engine

frisonisland:
I thought it could be a nice idea to add a tool that could help us test the API layer by making real calls.
This PR adds a MailboxSoapTest class that is meant to be extended in a test class.
You can run start and stop of SOAP server and make real calls with method executeSOAP.

**When starting the server you can register only required APIs.**
**The call makes it through ZimbraServlet logic and check for Authorization and everything**

- add MailboxSoapTest to allow easier testing of SOAP APIs
- change visibility of some methods/classes

### DIFF
--- a/store/src/main/java/com/zimbra/soap/SoapEngine.java
+++ b/store/src/main/java/com/zimbra/soap/SoapEngine.java
@@ -100,7 +100,7 @@ public class SoapEngine {
   private final DocumentDispatcher dispatcher = new DocumentDispatcher();
   private final MeterRegistry meterRegistry;
 
-  SoapEngine(MeterRegistry meterRegistry) {
+  public SoapEngine(MeterRegistry meterRegistry) {
     this.meterRegistry = meterRegistry;
     SoapTransport.setDefaultUserAgent(SoapTransport.DEFAULT_USER_AGENT_NAME, BuildInfo.VERSION);
   }
--- a/store/src/main/java/com/zimbra/soap/SoapEngine.java
+++ b/store/src/main/java/com/zimbra/soap/SoapEngine.java
@@ -100,7 +100,7 @@ public class SoapEngine {
   private final DocumentDispatcher dispatcher = new DocumentDispatcher();
   private final MeterRegistry meterRegistry;
 
-  public SoapEngine(MeterRegistry meterRegistry) {
+  SoapEngine(MeterRegistry meterRegistry) {
     this.meterRegistry = meterRegistry;
     SoapTransport.setDefaultUserAgent(SoapTransport.DEFAULT_USER_AGENT_NAME, BuildInfo.VERSION);
   }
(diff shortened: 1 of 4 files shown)


Polpetta:
Isn't really possible to achieve it without opening methods/constructors visibility for the production code?

frisonisland:
> Isn't really possible to achieve it without opening methods/constructors visibility for the production code?

Not really sure, as addService() is private, I need it to be at least protected, but I will check. Also there is a issue with this implementation: I would like to start/stop server in a static and non-static context (think BeforeAll vs BeforeEach). Having instance method covers only one scenario

frisonisland:
@Polpetta I made only addService of SoapServlet protected. Btw the whole idea of this class is to extend it and execute "executeSoapPost" to call a POST with SOAP body. 
It is meant to be very simple for now, just to get started with the idea of testing SOAP calls since the whole SOAP engine is custom.

Polpetta:
> Also there is a issue with this implementation: I would like to start/stop server in a static and non-static context (think BeforeAll vs BeforeEach). Having instance method covers only one scenario

We're gonna do it only and only if we need it. Otherwise it's a YagNi violation. We just have to implement the bare minimum caring about KISS.

On top of it, this class already adds logic into tests: how we can be sure this code itself works? It could contain bugs we don't see yet and this could lead to unexpected consequences. I'd do this kind of work once the necessity arises the first time imo

frisonisland:
> 

Agree, we can simply close this PR, however I hope my goal was clear: right now we are not able to run any SOAP call against our SOAP APIs. This PR adds a "test server" + utility class to allow this, through Jetty server. 
The test server is not mocked or anything, it is a real Jetty server.
Also:
- do we need to test test classes?
- By the time someone will use it to run a test, does that count as a test?
- what if someone adds the same class while developing a feature because it helps testing? Will it make sense then?

About starting/stopping I changed it to always do it before each/after each, also port is dynamic, I kept it minimal.

frisonisland:
> > Also there is a issue with this implementation: I would like to start/stop server in a static and non-static context (think BeforeAll vs BeforeEach). Having instance method covers only one scenario
> 
> We're gonna do it only and only if we need it. Otherwise it's a YagNi violation. We just have to implement the bare minimum caring about KISS.
> 
> On top of it, this class already adds logic into tests: how we can be sure this code itself works? It could contain bugs we don't see yet and this could lead to unexpected consequences. I'd do this kind of work once the necessity arises the first time imo


Agree, we can simply close this PR, however I hope my goal was clear: right now we are not able to run any SOAP call against our SOAP APIs. This PR adds a "test server" + utility class to allow this, through Jetty server. 
The test server is not mocked or anything, it is a real Jetty server.
Also:
- do we need to test test classes?
- By the time someone will use it to run a test, does that count as a test?
- what if someone adds the same class while developing a feature because it helps testing? Will it make sense then?

About starting/stopping I changed it to always do it before each/after each, also port is dynamic, I kept it minimal.

frisonisland:
Closing, will implement similar logic when in need

PR #320, opened 2023-09-06
chore: [CO-818] remove CSRF authentification

inno-dima:
**What has changed:**

- Removed CSRF filter from web.xml and web.xml.production
- Removed CSRF-related code
- Removed CSRF-related attributes


Polpetta (approved):

frisonisland:
Won't do for now, come back when testing is mature enough

PR #339, opened 2023-09-26
chore: 23.11.0 next devel cycle + build improvements

frisonisland:
This PR changes version from 23.10.0 to 23.11.0.
Also it adds a "modifier" property to add "-SNAPSHOT" on version.
If building with profile "prod" the modifier is empty, so the version becomes the release one.
Other changes:
- skip shellspec tests
- build with profile dev when on devel from the beginning.
- build with profile prod when on main from the beginning.


keshavbhatt (approved):
lgtm

aheeva-yuliya (approved):

PR #341, opened 2023-09-27
chore: modify ZimbraLog.slogger level to warn

aheeva-yuliya:
ref: CO-826

### DIFF
--- a/common/src/main/java/com/zimbra/common/stats/CsvStatsDumper.java
+++ b/common/src/main/java/com/zimbra/common/stats/CsvStatsDumper.java
@@ -109,7 +109,7 @@ public class CsvStatsDumper implements Dumper {
               + ","
               + line;
       if (logLine.length() <= SYSLOG_ELIDING_THRESHOLD) {
-        ZimbraLog.slogger.info(logLine);
+        ZimbraLog.slogger.warn(logLine);
       } else {
         StringBuilder b = new StringBuilder(logLine);
         String lastUuid = null;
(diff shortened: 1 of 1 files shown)


common/src/main/java/com/zimbra/common/stats/CsvStatsDumper.java
```diff
@@ -120,9 +120,9 @@ public Void call() throws Exception {
             sub = ":::" + lastUuid + ":::" + sub;
           }
           lastUuid = UUID.randomUUID().toString();
-          ZimbraLog.slogger.info(sub + ":::" + lastUuid + ":::");
+          ZimbraLog.slogger.warn(sub + ":::" + lastUuid + ":::");
         } while (b.length() > SYSLOG_ELIDING_THRESHOLD);
-        ZimbraLog.slogger.info(":::" + lastUuid + ":::" + b);
+        ZimbraLog.slogger.warn(":::" + lastUuid + ":::" + b);
```
keshavbhatt, line 125: I dont think this is sufficient, we need to update the log4j properties to define default log level. see, https://github.com/zextras/carbonio-mailbox/pull/237/files#diff-f6e59d703888489649baecee96e5baed5bb035ac4bb9add9d0fae414f4069808R209

keshavbhatt (changes requested):

common/src/main/java/com/zimbra/common/stats/CsvStatsDumper.java
frisonisland: Agree, this will not stop the logs to show. Did you check on a VM or locally with a test? (We should add a test)
We should raise log4j level to show only logs above INFO

frisonisland (changes requested):

keshavbhatt (approved):

inno-dima (approved):

frisonisland (approved):

PR #342, opened 2023-09-27
chore: add checkstyle

frisonisland:
- adds checkstyle in pre-commit


inno-dima (approved):

frisonisland:
By default checkstyle uses sun checks which is quite restrictive, we may want to use google checks. See: https://github.com/gherynos/pre-commit-java#notes-about-the-checkstyle-hook

PR #343, opened 2023-09-27
fix:  CO-823 FreeBusy missing uid parameter in CalDav request

frisonisland:
There is an open bug about the Free/Busy response missing the UID parameter.
From RFC https://www.rfc-editor.org/rfc/rfc5545 the FreeBusy component, VFREEBUSSY, is required to have the uid.

**Providing a null uid will result in the UID to not be added** as I don't know where to get the value in all other cases.
**But for this use case** (CalDAV request) and due to the current bug **I have decided to answer back the same UID that was provided in the request.**
Reason is while debugging with a client (Thunderbird), I saw the UID in the request was different every time, plus, quoting from the RFC:

>       When used to request free/busy time information, the "ATTENDEE"
>       property specifies the calendar users whose free/busy time is
>       being requested; the "ORGANIZER" property specifies the calendar
>       user who is requesting the free/busy time; the "DTSTART" and
>       "DTEND" properties specify the window of time for which the free/
>       busy time is being requested; the "UID" and "DTSTAMP" properties
>       are specified to assist in proper sequencing of multiple free/busy
>       time requests.
>

and

>       When used to reply to a request for free/busy time, the "ATTENDEE"
>       property specifies the calendar user responding to the free/busy
>       time request; the "ORGANIZER" property specifies the calendar user
>       that originally requested the free/busy time; the "FREEBUSY"
>       property specifies the free/busy time information (if it exists);
>       and the "UID" and "DTSTAMP" properties are specified to assist in
>       proper sequencing of multiple free/busy time replies.
>

About the tests: I have added a test about the response (behavioral)
and a couple of UT for the FreeBusy class. I didn't opt for a refactor or adding more tests, but this class is lacking a lot of them


store/src/test/java/com/zimbra/cs/dav/service/DavServletTest.java
```diff
@@ -171,4 +200,23 @@ void shouldDeleteAppointmentUsingCalDAV() throws Exception {
...
+  void shouldReturnUIDWhenRequestingFreeBusyOfAttendee() throws Exception {
+    Calendar calendar = new CalendarBuilder().build(this.getClass().getResourceAsStream("FreeBusyRequest_" + FREE_BUSY_UID + ".ics"), StandardCharsets.UTF_8.toString());
+
+    Assertions.assertEquals(
+        HttpStatus.SC_CREATED, createAppointmentWithCalDAV().getStatusLine().getStatusCode());
+    final HttpResponse response = requestAttendeeFreeBusy();
+    Assertions.assertEquals(
+        HttpStatus.SC_OK, response.getStatusLine().getStatusCode());
+    final String freeBusyResponse = new String(response.getEntity().getContent().readAllBytes());
+    System.out.println(freeBusyResponse);
```
inno-dima, line 219: Do we need this log?

inno-dima (approved):

keshavbhatt (approved):

store/src/test/java/com/zimbra/cs/dav/service/DavServletTest.java
frisonisland: Thx, I used it while testing and completely forgot about it :sweat_smile:

PR #358, opened 2023-10-18
chore: publish carbonio-mailbox-sdk client

matteobaglini:
Hi,

this PR completes the work started by @frisonisland, in particular:
- add a new `carbonio-mailbox-sdk` module
- auto generate the basic SOAP Client via `jaxws-maven-plugin`

In order to define it as a proper client SDK there is still a missing part of a "nice" client entry point which will be added in the future with a separate PR.


frisonisland (approved):
Ok for me, I see we opted for the "sdk" naming similarly to other projects

keshavbhatt (approved):

PR #367, opened 2023-10-30
test: local packages ce

frisonisland:
The Goal is to provide a way to test locally built packages in carbonio-ce in a convenient way.
Looking for suggestion/feedback also on file location/structure. Probably it would be nice to have some scripts to initialize some users (delegated admin, shared account, etc.) for common test scenarios


keshavbhatt (approved):

docker/single-server/entrypoint.sh
```diff
@@ -0,0 +1,26 @@
...
+#
+
+# Start services
+source /opt/zextras/.bashrc
+
+ldap start
+
+zmprov -l ms $(zmhostname) -zimbraServiceEnabled antivirus -zimbraServiceEnabled amavis -zimbraServiceEnabled stats
+zmprov -l mcf zimbraSmtpPort 25
+zmprov -l mcf zimbraSmtpHostname 127.0.0.1
```
keshavbhatt, line 15: we should merge these commands to gain little startup time. :wink:

docker/single-server/prepare_data.sh
```diff
@@ -0,0 +1,27 @@
...
+#
+# SPDX-FileCopyrightText: 2023 Zextras <https://www.zextras.com>
+#
+# SPDX-License-Identifier: AGPL-3.0-only
+#
+set -x
+
+cd /opt/zextras/bin || exit
+
+DOMAIN="carbonio-system.svc.cluster.local"
```
keshavbhatt, line 11: Can we change the domain to something simpler?
I heard from you that the default hostname is fixed for some reasons, but not sure its same for default domain.

docker/single-server/prepare_data.sh
```diff
@@ -0,0 +1,27 @@
...
+carbonio prov sp zextras password
+
+# Create test user
+carbonio prov ca test@"${DOMAIN}" password
+
+# Create admin user
+carbonio prov ca admin@"${DOMAIN}" password zimbraIsAdminAccount TRUE
+
+# Create shared account
+carbonio prov ca shared@"${DOMAIN}" password displayName "Shared"
```
keshavbhatt, line 22:
```suggestion
carbonio prov ca shared@"${DOMAIN}" password displayName "Shared account"
```
Just to avoid confusion in UI :wink:

aheeva-yuliya (approved):

PR #258, opened 2023-06-28
refactor(CO-764): remove 2FA, Devices, Trusted devices, App Specific Password, Remote Wipe

frisonisland:
This PRs removes 2FA logic and Devices + Trusted Devices and App Specific Passwords since it is not provided anymore.
**Plan to merge it for 23.8-23.9**

2FA:
- auth logic
- scratch codes

App specific password
- auth logic

Devices:
- verification code
- CRUD devices
- reminder device
- devices count
- sync

Trusted devices:
- auth logic
- token

Remote Wipe was part of Network Edition, so it was removed.
**The PR is quite big, but mostly because of deleted files.**
**I ask to please carefully check:**
- [admin Auth.java](https://github.com/zextras/carbonio-mailbox/pull/258/files#diff-7a16be07dac9a47a1769524350df860de80e4c1ece289a8da31103cbda60331e)
- https://github.com/zextras/carbonio-mailbox/pull/258/files#diff-7a16be07dac9a47a1769524350df860de80e4c1ece289a8da31103cbda60331e
- https://github.com/zextras/carbonio-mailbox/pull/258/files#diff-d79855035c7ea2d1dfe6889ee4796f0badea1aa28eac05f28792801778c908d3
- https://github.com/zextras/carbonio-mailbox/pull/258/files#diff-9d49b27d5fa0d6a57634e1382de08ffaa7fd2f6aa02d56b6d2e44be0925fcbf8
- https://github.com/zextras/carbonio-mailbox/pull/258/files#diff-b06e5582dd5e9c7544d9e7057d463c9242fb64fe9d6c4c8f46a92922b165a25a
- name of deleted request/response classes

**Manual testing**:
- zmsoap (was giving _ERROR: service.AUTH_REQUIRED (no valid authtoken present)_) fixed in https://github.com/zextras/carbonio-mailbox/pull/258/commits/b57cd85cf8b16a56b780a3a8d054e0e0763fd7d9
- IMAP (Thunderbird)

**TODO:**
- [x] cleanup https://github.com/zextras/carbonio-db-conf/blob/main/src/db/mysql/db.sql#L211 from devices

PR #372, opened 2023-11-06
[CO-877] Exclude NE endpoints from soap docs

sonersivri:


zextrasGiovanniDefacci (approved):

soap/src/main/java/com/zimbra/doc/soap/DocExcludedServices.java
```diff
@@ -0,0 +1,139 @@
...
+        SERVICE_SET.add("StartFailoverClient");
+        SERVICE_SET.add("StartFailoverDaemon");
+        SERVICE_SET.add("StopFailoverClient");
+        SERVICE_SET.add("StopFailoverDaemon");
+        SERVICE_SET.add("CancelPendingRemoteWipe");
+        SERVICE_SET.add("GetDeviceStatus");
+        SERVICE_SET.add("RemoteWipe");
+        SERVICE_SET.add("RemoveDevice");
+        SERVICE_SET.add("ResumeDevice");
+        SERVICE_SET.add("SuspendDevice");
```
frisonisland, line 113: It's ok for me to exclude them from a list, however in the past months we deleted some of these classes and they should not appear anymore in the SOAP doc. For example the RemoveDeviceRequest in https://github.com/zextras/carbonio-mailbox/pull/258/files and some others, so perhaps there is no need to exclude them?

lorenzopiovesanzextras (approved):

PR #381, opened 2023-11-14
[CO-862] fix: Autoprovisioning try to create account with domain

sonersivri:


frisonisland (changes requested):
I think we should add some tests to this if possible. Recently we approached a new way to do them by starting a  "real" mailbox, with an in memory ldap and database

sonersivri:
> I think we should add some tests to this if possible. Recently we approached a new way to do them by starting a "real" mailbox, with an in memory ldap and database

@frisonisland I added test

keshavbhatt (changes requested):
Suggested minor changes to improve maintianbility of code. Rest look okay to me.

keshavbhatt (approved):

frisonisland (approved):

PR #397, opened 2023-12-04
feat: [CO-917] generate regular auth token for recover password requests

keshavbhatt:
**What has changed:**
- Auth API now creates regular Auth token with AUTH usages when token is requested using password recovery code.


sonersivri (approved):

frisonisland (approved):
It's ok for me, although I would really like to use less of Zimbra static MailboxTestUtil in future :smile:

PR #400, opened 2023-12-06
feat: [CO-917] generate regular auth token for recover password requests (#397)

keshavbhatt:
**DO NOT SQUASH**


keshavbhatt:
It was decided to not release in 23.12.x but in 24.1.x and we already have it in 24.1.x devel.

PR #401, opened 2023-12-07
feat: healthcheck

frisonisland:
## Goal

The goal is to introduce the three health check endpoints (/health, /health/live, /health/ready) that should be available on every Carbonio service.

## Health check implementation
- Implemented endpoint/routing with HealthServlet
- Implemented HealthUseCase to aggregate readiness/liveliness based on different possible service healthchecks
- Implemented DatabaseServiceDependency to check database status. Has options to avoid keep polling connections (polling interval) and live + ready is based on the same check on connection availability.

## Testing
- added test containers dependency
- updated JettyServerFactory to allow setting filters and listeners
- Tested health check with Consul in production environment


sonersivri (approved):

store/src/main/java/com/zextras/mailbox/health/ServiceDependency.java
```diff
@@ -0,0 +1,61 @@
...
+public abstract class ServiceDependency implements HealthStatus {
+
+  private final String name;
+  private final ServiceType type;
+
+  private final int pollingIntervalMillis;
+  private final Supplier<Long> currentTimeProvider;
+  private Long lastCheckTimestamp;
+  private boolean lastHealthStatus;
+
```
frisonisland, line 21: @keshavbhatt I would avoid using these variable in the parent class. Reason is I would like to keep it as simple as possible. I would **maybe** create a PollingServiceDependency abstract class extending ServiceDependency. But even then there are some methods, like doCheckStatus() that migth not work for every use case. 
Since we currently have only one ServiceDependency I would think about doing this refactor later (YAGNI) and for now leave the polling logic just in DatabaseServiceDependency

keshavbhatt: I was about to add SoapServletServiceDependency, but i agree if there is no such plan we can skip this for now.

frisonisland: Ty, no rush, let's check it together when you are available :)

store/src/main/java/com/zextras/mailbox/servlet/HealthResponse.java
```diff
@@ -0,0 +1,116 @@
...
+   *   .build();
+   * }
+   * </pre>
+   */
+  static class HealthResponseBuilder {
+
+    private final List<Dependency> dependencies = new ArrayList<>();
+    private boolean ready = false;
+
+    private HealthResponseBuilder() {
```
ppolito-zextras, line 50: Why can't we leave the public constructor?

store/src/main/java/com/zextras/mailbox/servlet/HealthServlet.java
```diff
@@ -0,0 +1,76 @@
...
+  protected void doGet(HttpServletRequest httpServletRequest,
+      HttpServletResponse httpServletResponse)
+      throws ServletException, IOException {
+    String requestedPath = httpServletRequest.getPathInfo();
+    if (Objects.isNull(requestedPath)) {
+      requestedPath = "/";
+    }
+    final boolean isReady = healthUseCase.isReady();
+    switch (requestedPath) {
+      case "/":
```
ppolito-zextras, line 45: Talking about the IOException, the doGet method inherits Exceptions from the parent method. The try-catch block is not needed.

store/src/test/java/com/zextras/mailbox/health/DatabaseServiceDependencyTest.java
```diff
@@ -0,0 +1,140 @@
...
+        .thenThrow(ServiceException.FAILURE("Ooops, query failed"));
+    Mockito.when(currentTimeSupplier.get()).thenReturn(500L);
+    Assertions.assertTrue(databaseService.isReady());
+    Assertions.assertTrue(databaseService.isLive());
+  }
+
+
+  @Test
+  void shouldNotBeLiveAndReadyIfDbPoolConnectionCannotBeTaken()
+      throws ServiceException, SQLException {
```
ppolito-zextras, line 111: SQLException is never thrown in this method, you can cancel it

store/src/main/java/com/zextras/mailbox/servlet/HealthResponse.java
```diff
@@ -0,0 +1,116 @@
...
+
+  @SuppressWarnings("unused")
+  @JsonProperty("ready")
+  private final boolean ready;
+
+  @SuppressWarnings("unused")
+  @JsonProperty("dependencies")
+  private final List<Dependency> dependencies;
+
+  public HealthResponse(boolean ready, List<Dependency> dependencies) {
```
ppolito-zextras, line 27: Pay attention that Dependency is defined lower in this class so it's visible only for the classes of this package, but this method is visible to any class

frisonisland: Yes, IMHO there is no need to have them package-private, unless we want to enforce strict rules. Having them public won't hurt, same for private constructor and enforcing usage of static method newInstance()

store/src/test/java/com/zextras/mailbox/util/JettyServerFactory.java
```diff
@@ -4,18 +4,26 @@
...
+import javax.servlet.DispatcherType;
 import org.eclipse.jetty.server.Connector;
 import org.eclipse.jetty.server.Server;
 import org.eclipse.jetty.server.ServerConnector;
+import org.eclipse.jetty.servlet.FilterHolder;
 import org.eclipse.jetty.servlet.ServletContextHandler;
 import org.eclipse.jetty.servlet.ServletHolder;
 
 /** Test utility class to create a {@link Server} instance with custom port and servlets. */
 public class JettyServerFactory {
```
ppolito-zextras, line 23: The following variables can be set as final (servlets, filters, listeners and host)

store/src/main/java/com/zextras/mailbox/servlet/HealthServlet.java
keshavbhatt: It is considered a good practice to not throw exception in Servlet methods. 
see: https://rules.sonarsource.com/java/RSPEC-1989/

store/src/main/java/com/zextras/mailbox/servlet/HealthResponse.java
keshavbhatt: It can be. The idea is not to worry about how to create the builder itself + make the usages more expressive and readable.

store/src/main/java/com/zextras/mailbox/servlet/HealthServlet.java
frisonisland: Even if it is a good practice the issue is the IOException is already defined in the method.
At that point I would remove that. Same for others, else we should just do a big try{} catch{}

store/src/main/java/com/zextras/mailbox/health/ServiceDependency.java
frisonisland: I've put back the old behavior

packages/appserver-service/carbonio-mailbox.hcl
```diff
@@ -1,6 +1,6 @@
 services {
   check {
-    tcp      = "localhost:8080"
+    http      = "http://localhost:8080/service/health/ready"
```
keshavbhatt, line 3: I this it is premature to have this change at the moment. 
The rationale is that the current health endpoint readiness only reflects the readiness of the database, which in itself does not indicate the overall health of the mailbox.

keshavbhatt (changes requested):
I just ask if it is okay to rely on this endpoint in production to suggest if mailbox is healthy. I suggest to wait for addition of other required service dependencies.

keshavbhatt (approved):

packages/appserver-service/carbonio-mailbox.hcl
frisonisland: Sorry if it sounded harsh, was not my intention at all, but I think it is a pity not use the endpoint + it's already an improvement over the original behavior. Also it will pass quite some time before we reach a fully mature health-check. By using it we can gain some feedback on its performance already + possible issues

matteobaglini (approved):

PR #416, opened 2023-12-28
Fix (CO-933): make SearchDirectory response isMore and searchTotal fields properly work

AlessioCoser:
We fixed the behaviour of SearchDirectory command:
- now the 'more' field that tell if there are more pages other than the current is working
- now the 'searchTotal' field provides the correct total items

The current behaviour mimics the one present in the GetQuotaUsage command


matteobaglini (approved):

danielemegna (approved):
:rocket:

PR #425, opened 2024-01-10
chore: devel cycle 24.2.0-SNAPSHOT + pkgver 4.13.0

keshavbhatt:


sonersivri (approved):

frisonisland:
We should update also the version.txt for release-it if we want to still use it, else we can consider removing it. 
However this will result in having a missing unreleased package version every time, because:
- start from 4.12.0
- bump manually to 4.13.0
- bump with release-it to 4.14.0 or 4.13.1, depending on changes, on release

At this point I would consider removing release-it and do it manually every time or use another strategy

aheeva-yuliya (approved):

PR #428, opened 2024-01-11
refactor: [CO-949] - move and load rights from resources

ppolito-zextras:
- refactor RightManager to use AttributeManager
- move the rights files to resources
- made the RightManager load the rights from the resources

TODO:
- [x] test against advanced Integration Tests


store/src/main/java/com/zimbra/cs/account/accesscontrol/RightManager.java
```diff
@@ -122,73 +126,109 @@ static String listCoreDefFiles() {
...
+        @Override
+        public InputStream getContent(String rightFileName) throws ServiceException {
+            final File file = new File(baseDirectory, rightFileName);
+            try {
+                return new FileInputStream(file);
+            } catch (FileNotFoundException e) {
+                throw ServiceException.FAILURE("Cannot read file " + file.getAbsolutePath(), e);
             }
+        }
+    }
```
frisonisland, line 181: Move these classes out. They are not totally related to Rights, as reading content from a file or resource doesnn't have anything to do with rights.
Example: you can do new ResourceContentExtractor("/conf/rights") -> this could allow reading resources from the base path "/conf/rights".
Similarly, new FileContentExtractor("/conf/rights") -> reads from file system in base directory "/conf/rights".

store/src/test/java/com/zimbra/cs/account/accesscontrol/RightManagerTest.java
```diff
@@ -0,0 +1,56 @@
...
+    final AttributeManager attributeManager = new AttributeManager("../store/conf/attrs");
+
+    final ServiceException thrownException = Assertions.assertThrows(ServiceException.class,
+        () -> RightManager.fromFileSystem("src/main/resources/non-existing", attributeManager));
+    Assertions.assertEquals("system failure: rights directory does not exist: src/main/resources/non-existing", thrownException.getMessage());
+  }
+
+  private void assertRightsLoaded(RightManager rightManager) throws ServiceException {
+    Assertions.assertFalse(rightManager.getAllAdminRights().isEmpty());
+    Assertions.assertFalse(rightManager.getAllUserRights().isEmpty());
```
frisonisland, line 47: Not needed, below assertions already include these checks

store/src/test/java/com/zimbra/cs/account/accesscontrol/RightManagerTest.java
```diff
@@ -0,0 +1,56 @@
...
+  @Test
+  void shouldLoadRightsUsingResources() throws ServiceException {
+    final AttributeManager attributeManager = new AttributeManager("../store/conf/attrs");
+    RightManager rightManager = RightManager.fromResources(attributeManager);
+
+    assertRightsLoaded(rightManager);
+  }
+
+  @Test
+  void shouldLoadRightsUsingSingleton() throws ServiceException {
```
frisonisland, line 29: Since this test is does not set zimbra_rights_directory, and value is empty by default, it uses resources.
Rename method to something like _shouldLoadRightsfromResourcesUsingSingleton_

store/src/test/java/com/zimbra/cs/account/accesscontrol/RightManagerTest.java
```diff
@@ -0,0 +1,56 @@
+package com.zimbra.cs.account.accesscontrol;
+
+import com.zimbra.common.localconfig.LC;
+import com.zimbra.common.service.ServiceException;
+import com.zimbra.cs.account.AttributeManager;
+import org.junit.jupiter.api.Assertions;
+import org.junit.jupiter.api.Test;
+
+class RightManagerTest {
```
frisonisland, line 9: We should add other singleton tests to check it works using filesystem.
However we must remember to destroy the singleton at the end of the test + reset the LC.zimbra_rights_directory.

frisonisland (changes requested):

store/src/main/java/com/zimbra/cs/account/accesscontrol/RightManager.java
frisonisland: Decided to not do this and opt for another solution

keshavbhatt (approved):

frisonisland (approved):

PR #433, opened 2024-01-22
chore(docker): move from ldap-server to carbonio-ce-directory-server

frisonisland:
- image was renamed
- updated checks
- updated hostname to match new one defined during bootstrap

### DIFF
--- a/docker/jetty-run/docker-compose.yml
+++ b/docker/jetty-run/docker-compose.yml
@@ -28,10 +28,10 @@ services:
       - "~/.m2:/root/.m2"
 
   ldap:
-    image: carbonio/ce-ldap-u20
-    hostname: ldap.mail.local
+    image: carbonio/ce-directory-server-u20:23.12.0
+    hostname: carbonio-ce-directory-server.carbonio-system.svc.cluster.local
     healthcheck:
-      test: nc -z localhost 2812 || exit 1 # check monit up and running
+      test: netstat -ltn | grep -c :389
       retries: 20
     ports:
       - "389:389"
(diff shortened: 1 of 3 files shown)


sonersivri (approved):

docker/jetty-run/docker-compose.yml
```diff
@@ -28,17 +28,18 @@ services:
       - "~/.m2:/root/.m2"
 
   ldap:
-    image: carbonio/ce-ldap-u20
-    hostname: ldap.mail.local
+    image: carbonio/ce-directory-server-u20:23.12.0
+    hostname: carbonio-ce-directory-server.carbonio-system.svc.cluster.local
     healthcheck:
-      test: nc -z localhost 2812 || exit 1 # check monit up and running
+      test: netstat -ltn | grep -c :389
```
M0Rf30, line 34: you can use 
nc -z localhost 389 || exit 1

frisonisland: > you can use nc -z localhost 389 || exit 1

I don't think this works anymore as the process is not listening on loopback. As a matter of fact the app does not start as it depends on ldap which result in a failing healthcheck.
I decided to use 'netstat -ltn | grep -c :389' because it returns matches found, else 0. In first case (match, 1) exit status is 0, in the other case is 1

M0Rf30: Just as extratip:
nc want hostname and port. having a static declared and fixed hostname you can also do:
```
nc -z carbonio.carbonio-system.svc.cluster.local 389 || exit 1
```
Humble suggestion

keshavbhatt (approved):

docker/jetty-run/entrypoint-ldap.sh
M0Rf30, line 1: align `carbonio/ce-ldap-u20` in the comment on line 3 with
`carbonio/ce-directory-server-u20`

can I also ask why do you need the
`touch /.dockerenv` 
instruction?

sonersivri (approved):

PR #435, opened 2024-01-23
fix: [CO-862] Autoprovisioning zimbraAutoProvAccountNameMap account create

sonersivri:


mbarto (approved):

store/src/main/resources/conf/attrs/attrs.xml
```diff
@@ -10128,4 +10128,9 @@ TODO: delete them permanently from here
   <globalConfigValue>yyyyMMddHHmmss.SSS'Z'</globalConfigValue>
   <desc>AD/LDAP timestamp format. For example: yyyyMMddHHmmss.SSS'Z', yyyyMMddHHmmss'Z'</desc>
 </attr>
+<attr id="3142" name="carbonioAutoProvAllowedDomains" type="string" max="51200" cardinality="single" optionalIn="domain" since="24.2.0">
+  <desc>
+    carbonioAutoProvAllowedDomains comma separated domain names
```
keshavbhatt, line 10133: This description for the attrubute can be improved. 
We can remove the name in the description since it is redundant. Also, adding the purpose of the attribute rather than just providing the usages in description can be helpful.

keshavbhatt (changes requested):

store/src/main/java/com/zimbra/cs/account/ldap/AutoProvision.java
```diff
@@ -277,23 +277,34 @@ protected String mapName(ZAttributes externalAttrs, String loginName)
...
-                        }
-                        break;
-                    case SEARCH_BY_USER_PRINCIPAL_NAME:
-                        localPart = localPartArr[EMAIL_INDEX];
-                        break;
+                String attrDomain = localPartArr[DOMAIN_INDEX].trim();
+                if (domain.getName().equals(attrDomain) || getDomainSet().contains(attrDomain)) {
+                    localPart = localPartArr[EMAIL_INDEX];
+                } else {
+                    throw ServiceException.FAILURE(localPart + " can not be provisioned for domain " + domain.getName());
```
keshavbhatt, line 284: Can we make this exception more helpful, putting keywords like "not allowed", "should not contain" or something that suggest why this failure occured?

store/src/main/resources/conf/attrs/attrs.xml
sonersivri: @keshavbhatt I changed could you check?

store/src/main/java/com/zimbra/cs/account/ldap/AutoProvision.java
```diff
@@ -277,23 +277,34 @@ protected String mapName(ZAttributes externalAttrs, String loginName)
...
                 }
             }
         }
+
         return localPart + AT_SIGN + domain.getName();
     }
 
+    Set<String> getDomainSet() {
```
keshavbhatt, line 292: This method name `getDomainSet` can be improved. We can make it util method if it is doing generic things maybe passing domain as parameter? What do you think?

sonersivri: changed name of the method getCarbonioAutoProvAllowedDomainSet

sonersivri: @keshavbhatt

keshavbhatt (approved):

store/src/main/resources/conf/attrs/attrs.xml
keshavbhatt: @sonersivri I think we also need to regenrate the getters and setters to update the description in javadocs as well :)

PR #444, opened 2024-02-15
refactor: [CO-983] provutil

frisonisland:
The main goal is to introduce some tests on ProvUtil so we can test the CLI without deploying.
These tests require a SOAP service to be up so the CLI can interact with it.

We changed ProvUtil system.out/err redirection and refactored to use a Console class for printing the output, so we can intercept the content and write some assertions.


aheeva-yuliya (approved):

store/src/test/java/com/zimbra/cs/account/ProvUtilTest.java
```diff
@@ -0,0 +1,445 @@
...
+      .create();
+
+  @BeforeAll
+  static void setUp() {
+    //noinspection HttpUrlsUsage
+    LC.zimbra_admin_service_scheme.setDefault("http://");
+    LC.zimbra_admin_service_port.setDefault(SOAP_PORT);
+  }
+
+  @AfterAll
```
AlessioCoser, line 50: Why do we need an init in the AfterAll section?

mbarto (approved):

keshavbhatt (approved):
I thought we will put some work here but its okay.

AlessioCoser (approved):

store/src/main/java/com/zimbra/cs/account/ProvUtil.java
```diff
@@ -4078,7 +3929,7 @@ private void checkDeprecatedAttrs(Map<String, ? extends Object> attrs) throws Se
     }
 
     if (hadWarnings) {
-      console.println();
+      console.print("\n");
```
matteobaglini, line 3932: Hi @frisonisland, it's minor thing but I think replace `console.print("\n");` with `console.println();` make the api more familiar.

store/src/main/java/com/zimbra/cs/account/ProvUtil.java
```diff
@@ -1345,8 +1171,26 @@ private boolean needProvisioningInstance(Command command) {
...
+      if (commandMap.get(name) != null) {
+        throw new RuntimeException("duplicate command: " + name);
+      }
+      String alias = c.getAlias().toLowerCase();
+      if (commandMap.get(alias) != null) {
+        throw new RuntimeException("duplicate command: " + alias);
+      }
+      commandMap.put(name, c);
+      commandMap.put(alias, c);
+    }
```
matteobaglini, line 1187: Hi @frisonisland, here we can extract a function `getCommandMap` and place it in `Command` class.

store/src/main/java/com/zimbra/cs/account/ProvUtil.java
```diff
@@ -2129,12 +1973,12 @@ private void doListHabOrgUnit(String[] args) throws ServiceException {
     Domain domain = lookupDomain(args[1], prov, Boolean.FALSE);
     Set<String> resultSet;
     if (prov instanceof SoapProvisioning) {
-      resultSet = ((SoapProvisioning) prov).listHabOrgUnit(domain);
+      resultSet = prov.listHabOrgUnit(domain);
```
matteobaglini, line 1976: Hi @frisonisland, now the two if branches are equals, can we keep only one?

store/src/main/java/com/zimbra/cs/account/ProvUtil.java
```diff
@@ -2146,7 +1990,7 @@ private void doRenameHabOrgUnit(String[] args) throws ServiceException {
     }
     Domain domain = lookupDomain(args[1], prov, Boolean.FALSE);
     if (prov instanceof SoapProvisioning) {
-      ((SoapProvisioning) prov).renameHabOrgUnit(domain, args[2], args[3]);
+      prov.renameHabOrgUnit(domain, args[2], args[3]);
```
matteobaglini, line 1993: Hi @frisonisland , here same idea as previous comment.

store/src/main/java/com/zimbra/cs/account/ProvUtil.java
```diff
@@ -2159,7 +2003,7 @@ private void doDeleteHabOrgUnit(String[] args) throws ServiceException {
     }
     Domain domain = lookupDomain(args[1], prov, Boolean.FALSE);
     if (prov instanceof SoapProvisioning) {
-      ((SoapProvisioning) prov).deleteHabOrgUnit(domain, args[2]);
+      prov.deleteHabOrgUnit(domain, args[2]);
```
matteobaglini, line 2006: Same :-)

store/src/test/java/com/zimbra/cs/account/ProvUtilTest.java
```diff
@@ -0,0 +1,445 @@
...
+
+    final String deleteOperationResult = runCommand(new String[]{"da", accountName});
+    Assertions.assertTrue(deleteOperationResult.isEmpty());
+
+    OutputStream outputStream = new ByteArrayOutputStream();
+    OutputStream errorStream = new ByteArrayOutputStream();
+    catchSystemExit(() -> runCommand(outputStream, errorStream, new String[]{"ga", accountName}));
+    final String expectedError =
+        "ERROR: account.NO_SUCH_ACCOUNT (no such account: " + accountName + ")\n";
+    Assertions.assertEquals(expectedError, errorStream.toString());
```
matteobaglini, line 209: Hi @frisonisland, in this case I prefer a more relaxed assert something like `assertStringContains("NO_SUCH_ACCOUNT", errorStream.toString());`. So little modification to the error message do not produce a lot of red tests.
It's my personal taste so even `assertEquals` is ok.

matteobaglini (approved):

store/src/main/java/com/zimbra/cs/account/ProvUtil.java
frisonisland: Done!

store/src/main/java/com/zimbra/cs/account/ProvUtil.java
frisonisland: Done, found other similar cases

store/src/test/java/com/zimbra/cs/account/ProvUtilTest.java
frisonisland: I'll keep it as is for now, I don't think messages will change much, let's see in the future :stuck_out_tongue_winking_eye:

PR #450, opened 2024-02-20
feat(CO-954): define Mailbox Admin Sidecar

matteobaglini:


packages/appserver-service/carbonio-mailbox-admin.hcl
```diff
@@ -0,0 +1,17 @@
+services {
+  check {
+    http      = "http://localhost:8080/service/health/ready"
```
frisonisland, line 3: Is this endpoint available also on port 7071? 8080 is technically non-admin service

matteobaglini: Ok, let me check

frisonisland (approved):

packages/appserver-service/carbonio-mailbox-admin.hcl
matteobaglini: Hi @frisonisland, we checked and in fact the same health endpoint also works with the for admin (on 7071). Fixed!

PR #451, opened 2024-02-20
chore: [CO-988] use project mvn revision property to generate build info class

keshavbhatt:
**What has changed:**

Project revision property set in project `pom.xml` is shared with ant task that generate carbonio-build-info(`BuildInfoGenerated`) classes.

This omits the need to update the `ant-build-common.xml` during release process.

**Bonus:**

- Added SOAP API changelog generation ant tasks. 
    - `generate-soap-api-changelog` to generate changelog comparing the baseline api description json.
    - `update-baseline-apidesc` to update the existing baseline api description json
    - Updated the changelog generation template and style.

- Simplification and cleanup of related code and instrumentation.


ant-build-common.xml
```diff
@@ -62,8 +62,9 @@
     <echo level="info" message="Using version ${zimbra.buildinfo.version}"/>
   </target>
 
+  <property file="${maven.build.dir}/mvn-ant-shared-build.properties"/>
   <target name="carbonio-version">
-    <property name="carbonio.buildinfo.version" value="24.4.0_ZEXTRAS_202404"/>
+    <property name="carbonio.buildinfo.version" value="${mvn_revision}"/>
```
frisonisland, line 67: Just a question: what is the value of mvn_revision?
I see that mvn-ant-shared-build.properties is generated with mvn_revision value equal to revision, but revision is 24.4.0 and we miss the _ZEXTRAS_202404 part. Maybe we do not need it anymore?

keshavbhatt: It is needed, it is used as build timestamp in generated build info. 
I provided alternative solution that uses current datetime YYYYMM format.

store/src/main/java/com/zimbra/cs/util/BuildInfo.java
```diff
@@ -32,15 +31,18 @@ public final class BuildInfo {
 
     public static final String FULL_VERSION;
 
+    public static final String UNKNOWN = "unknown";
```
frisonisland, line 34: Just a suggestion that we can make this private to avoid exposing things outside

keshavbhatt: sure

matteobaglini (approved):

PR #458, opened 2024-03-14
chore: Exclude generated classes from SonarQube analysis

keshavbhatt:
**Motivation:**
We've noticed that during SonarQube scanning, certain autogenerated classes(`**/com/zimbra/cs/account/ZAttr*.java,**/com/zimbra/common/account/ZAttr*.java`) are being included in the analysis. These classes typically contain thousands of lines of Java code and consist mainly of simple getter, setter, and unsetter methods. As these classes are autogenerated and not manually maintained, they don't require scrutiny from SonarQube, and their inclusion in the analysis results in noise and unnecessary reports.

**In this MR:**
We update the CI to exclude these autogenerated classes from SonarQube scanning process. By doing so, we can focus SonarQube's analysis on the code that requires attention and avoid false positives.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -153,7 +153,7 @@ pipeline {
             }
             steps {
                 withSonarQubeEnv(credentialsId: 'sonarqube-user-token', installationName: 'SonarQube instance') {
-                    mvnCmd("$BUILD_PROPERTIES_PARAMS sonar:sonar -Dsonar.exclusions=**/ZAttr*.java")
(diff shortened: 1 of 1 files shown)


frisonisland (approved):

sonersivri (approved):

M0Rf30 (commented):
Seems that it's indexing 0 files :(

keshavbhatt:
> Seems that it's indexing 0 files :(

Yes, since i did not touched any files in this PR other than Jenkins :) 
I tested with a simulation that the syntax we put in place works.

PR #464, opened 2024-03-19
chore: fix CI

keshavbhatt:
This change ensure that we use same version of JDK/JVM(11) across all stages of our pipeline.

We have two versions (11 and 17) of JDK and java runtime in our build environment/agent  since 94c777469a2d24ca5fafa1c7a6bf056ae64f7367 due to that some of the statges were failing.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -55,18 +55,23 @@ pipeline {
             label 'carbonio-agent-v2'
         }
     }
+
     parameters {
         booleanParam defaultValue: false, description: 'Upload packages in playground repositories.', name: 'PLAYGROUND'
         booleanParam defaultValue: false, description: 'Skip test and sonar analysis.', name: 'SKIP_TEST_WITH_COVERAGE'
         booleanParam defaultValue: false, description: 'Skip sonar analysis.', name: 'SKIP_SONARQUBE'
     }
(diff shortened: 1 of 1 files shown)


M0Rf30:
This will be valid for all the stages, That's why I selected only the ones that really need jdk11.
If propagated to sonarqube stage, for example, it will trigger it using jdk11 ang generating also a warning.

keshavbhatt:
> This will be valid for all the stages, That's why I selected only the ones that really need jdk11. If propagated to sonarqube stage, for example, it will trigger it using jdk11 ang generating also a warning.

So, for sonarqube analysis stage we can use JDK17?  Feel free to commit on this branch.
Thanks for clarifying.

sonersivri (approved):

aheeva-yuliya (approved):

M0Rf30:
So in this way also sonarlint is executed with jdk11 I think

keshavbhatt:
> So in this way also sonarlint is executed with jdk11 I think

With this change, we override the JDK/JVM to version 17 for sonarqube anaysis stage. So, we assume the sonarqube analysis is using JDK/JVM 17. To confirm this we checked the warning int he dashboard and it is no more.

PR #473, opened 2024-04-04
feat: [CO-982] Enhance ordering of matches in FullAutoComplete

keshavbhatt:
**What has changed:**
- A new element `'orderedAccountIds'` was introduced in `FullAutocompleteRequest` . This field is used to specify the accounts from which autocomplete matches will be returned. The order of the account IDs in this field determines the  order in which the macthes are returned in `FullAutocompleteResponse` .
- Enhacned ordering algorithm to return FullAutocomplete matches. The new sorting/ordering and limiting algorithm works as follows:
    - Matches from the most preferred source account (the first accountId in 'orderedAccountIds') are given higher priority and are placed on top.
    - Matches from other source accounts are sorted primarily based on ranking, followed by alphabetical ordering.
    - Duplicate matches are omitted keeping the highest ranked match from both preferred account and other source accounts.
    - If the optional `'orderedAccountIds'` element is not passed, matches from the authenticated Account are returned.
    - The number of matches returned depends on the value set for authenticated account's `ContactAutoCompleteMaxResults` attribute.
    
- API call no more fails if call to retrieve matches from one or more  remote accounts fails due to permissions, network issues, or for any other reason. Instead, the failure is logged, and matches from the functioning accounts/mailboxes are returned.

### DIFF
--- a/store/src/main/java/com/zimbra/cs/service/mail/FullAutoComplete.java
+++ b/store/src/main/java/com/zimbra/cs/service/mail/FullAutoComplete.java
@@ -62,14 +62,13 @@ public class FullAutoComplete extends MailDocumentHandler {
               .filter(autoCompleteMatch -> fullAutoCompleteMatches.stream()
                   .noneMatch(m -> m.getEmail().equalsIgnoreCase(autoCompleteMatch.getEmail())))
               .forEachOrdered(otherAutoCompleteMatches::add));
-
-      otherAutoCompleteMatches.sort(Comparator.comparing(AutoCompleteMatch::getRanking).reversed()
-          .thenComparing(AutoCompleteMatch::getEmail));
     } catch (ServiceException e) {
       throw ServiceException.FAILURE(e.getMessage());
     }
 
     otherAutoCompleteMatches.stream()
+        .sorted(Comparator.comparing(AutoCompleteMatch::getRanking).reversed()
+            .thenComparing(AutoCompleteMatch::getEmail))
         .filter(autoCompleteMatch -> fullAutoCompleteMatches.stream()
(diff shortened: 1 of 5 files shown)


sonersivri (approved):

aheeva-yuliya (approved):

frisonisland (approved):

store/src/main/java/com/zimbra/cs/service/mail/FullAutoComplete.java
```diff
@@ -8,92 +8,167 @@
...
+      final var parsedAccountIds = parsePreferredAccountsFrom(
+          fullAutocompleteRequest.getOrderedAccountIds());
+      final var preferredAccountId = parsedAccountIds._1();
+      final var otherPreferredAccountIds = parsedAccountIds._2();
+
+      doAutoCompleteOnAccount(authenticatedAccount, zAuthToken, preferredAccountId, autoCompleteRequest)
+          .getMatches()
+          .stream()
+          .limit(Math.max(contactAutoCompleteMaxResultsLimit, 0))
+          .forEachOrdered(fullAutoCompleteMatches::add);
```
mbarto, line 56: What do you think of refactoring this to return a collection instead of adding items as a side effect in forEachOrdered?

mbarto (approved):
looks good to me, just a suggestion in the code (ignorable)

store/src/main/java/com/zimbra/cs/service/mail/FullAutoComplete.java
keshavbhatt: I agree, The idea here was to perform the action in one call(single steam pipeline), avoiding additional computation in a seperate `ArrayList.addAll`.

PR #477, opened 2024-04-08
feat: add systemd units

M0Rf30:
Refs: IN-557


sonersivri (approved):

frisonisland:
Hi @M0Rf30 what about adding systemd units for mailbox and milter on https://github.com/zextras/carbonio-appserver/blob/main/appserver/PKGBUILD ? Currently appserver-service contains mailbox sidecar and I'm not sure about putting milter service in carbonio-common-appserver-conf

keshavbhatt (approved):
Aside from the placement of these files, as noted by @frisonisland, everything appears satisfactory. :rocket:

frisonisland (approved):

PR #490, opened 2024-04-15
refactor: [CO-1046] embedded jetty

frisonisland:
Adds support for embedded jetty with same customization as defined in https://github.com/zextras/carbonio-appserver/blob/main/appserver/conf/jetty/jetty.xml.production.

**See:** https://github.com/zextras/carbonio-core/tree/refactor/CO-1046

Benefits:
- can start mailbox locally with same configuration as in production
- full control over startup process
- simplifies deployment
- removes risk of LinkageError caused by common dependencies used in Jetty startup and application

## Changes
- Mailbox.java (entrypoint)
- LikeJettyXML class to setup JettyServer as jetty.xml config
- rename web.xml -> web-dev.xml,  web.xml.production -> web.xml
- define store module as jar (not war anymore)
- add jython-libs module (separate jython only dependencies)
- add /usr/bin/carbonio-appserver to start mailbox (Mailbox.java). Used by system unit and zmmailboxdmgr
- put mailbox jars under /opt/zextras/mailbox/jars instead of jetty_base (no longer supported)
- **remove packages** carbonio-common-core-libs, carbonio-common-core-jar, carbonio-appserver-store-libs, carbonio-appserver-war in favor of carbonio-mailbox-jar (include store fat-jar + carbonio-jetty-libs dependencies). 

### Test
- add main class SampleLocalMailbox.java with in-memory databases for local testing
- add docker multi-server to test changes (Only LDAP + Proxy + Appserver)


packages/directory-server/PKGBUILD
```diff
@@ -110,52 +110,3 @@ postinst__rocky_8() {
     su - zextras -c "/opt/zextras/bin/ldap restart"
   fi
 }
-
-postinst__rocky_9() {
```
M0Rf30, line 114: Why?
there are some different behaviours between rhel8 and rhel9

packages/mailbox-jar/PKGBUILD
```diff
@@ -0,0 +1,33 @@
+pkgname="carbonio-mailbox-jar"
+pkgver="4.16.0"
+pkgrel="1"
+pkgdesc="Carbonio Mailbox Jars"
```
M0Rf30, line 4: is this new package really needed?

M0Rf30 (changes requested):

store/conf/web.xml
```diff
@@ -1,16 +1,14 @@
...
-SPDX-FileCopyrightText: 2022 Zextras <https://www.zextras.com>
-
-SPDX-License-Identifier: GPL-2.0-only
--->
+  ~ SPDX-FileCopyrightText: 2024 Zextras <https://www.zextras.com>
+  ~
+  ~ SPDX-License-Identifier: AGPL-3.0-only
+  -->
 
 <web-app xmlns="http://java.sun.com/xml/ns/j2ee"
```
keshavbhatt, line 8: Hi @frisonisland, is this file intended to be used in production? I am asking since now i see 3 files
- web.xml
- web-dev.xml
- web.xml.production

store/pom.xml
```diff
@@ -533,12 +533,32 @@
       <version>1.19.0</version>
       <scope>test</scope>
     </dependency>
+    <dependency>
+      <groupId>org.eclipse.jetty</groupId>
+      <artifactId>jetty-webapp</artifactId>
+      <version>9.4.48.v20220622</version>
```
keshavbhatt, line 539: @frisonisland  This also means we do not need `carbonio-jetty-distribution` anymore right?

keshavbhatt (approved):

store/conf/web.xml
frisonisland: Thanks, I will check it again. This branch is so long-lived that I can't recall why this file is here

frisonisland: Ok, I found it is used by the jetty-maven-plugin for local development purposes. However this file was always there in the repo since before we introduced the plugin.
I think if we manage to merge this PR then we can get rid of maven jetty plugin, since you can run the mailbox with a Main class, and maybe get rid of this web.xml.

packages/directory-server/PKGBUILD
frisonisland: Updated and also aligned appserver.service file. There are no changes regarding directory-server pkgbuild anymore

packages/mailbox-jar/PKGBUILD
frisonisland: We discussed and the reason is this package ships all mailbox jars. It is different from "appserver".
Appserver is a package that "ships" these jars + service definition + other things to make a whole node mailbox. Appserver also might not be limited to the mailbox

store/pom.xml
frisonisland: Correct, carbonio-jetty-distribution is part of carbonio-thirds if I'm not mistaken and can be removed. Even if installed, it will not cause any troubles since it will be unused and the new jar classpath will be different from the Jetty one.

build_packages.sh
```diff
@@ -4,11 +4,30 @@
...
 #
+OS=$1
 
-docker run -it --rm \
-  --entrypoint=yap \
-  -v $(pwd)/artifacts:/artifacts \
-  -v $(pwd):/tmp/staging \
-  docker.io/m0rf30/yap-ubuntu-focal:1.6 \
-  build ubuntu-focal /tmp/staging/packages
+if [[ -z $OS ]]
```
matteobaglini, line 9: Hi @frisonisland, how about providing a default value for the OS instead of blocking execution? In the end most of our VMs are all Ubuntu. What do you think?

frisonisland: Agree

store/src/main/java/com/zextras/mailbox/LikeXmlJettyServer.java
```diff
@@ -0,0 +1,273 @@
...
+import org.eclipse.jetty.server.handler.HandlerCollection;
+import org.eclipse.jetty.server.handler.RequestLogHandler;
+import org.eclipse.jetty.server.handler.gzip.GzipHandler;
+import org.eclipse.jetty.util.ssl.SslContextFactory;
+import org.eclipse.jetty.util.thread.QueuedThreadPool;
+import org.eclipse.jetty.util.thread.ThreadPool;
+import org.eclipse.jetty.webapp.WebAppContext;
+
+// See: https://github.com/jetty/jetty.project/blob/jetty-9.4.x/examples/embedded/src/main/java/org/eclipse/jetty/embedded/LikeJettyXml.java
+// Plus we adopted the same variables used in https://github.com/zextras/carbonio-appserver/blob/main/appserver/conf/jetty/jetty.xml.production
```
matteobaglini, line 37: Hi @frisonisland, minor note: if I'm not wrong this link will be broken once this PR is merged and released. Better to point to a specific tag instead of main, right?

store/src/test/java/com/zextras/mailbox/LikeXmlJettyServerTest.java
```diff
@@ -0,0 +1,59 @@
...
+import com.zextras.mailbox.LikeXmlJettyServer.InstantiationException;
+import com.zimbra.cs.account.Config;
+import org.eclipse.jetty.server.Connector;
+import org.eclipse.jetty.server.Server;
+import org.junit.jupiter.api.Test;
+
+class LikeXmlJettyServerTest {
+
+  @Test
+  void shouldCreateJettyServer() throws InstantiationException {
```
matteobaglini, line 21: Hi @frisonisland , this test is big and generic. We could divide it by groups of assertions: a test for userConnector, another for adminConnector, etc. and centralize the mock setup part.

matteobaglini (approved):

store/src/test/java/com/zextras/mailbox/SampleLocalMailbox.java
```diff
@@ -0,0 +1,103 @@
...
+    LC.mailboxd_keystore.setDefault(keystorePath.toAbsolutePath().toString());
+    LC.mailboxd_keystore.setDefault(keystorePath.toAbsolutePath().toString());
+    LC.zimbra_server_hostname.setDefault(APP_SERVER_NAME);
+    LC.timezone_file.setDefault("store/src/test/resources/timezones-test.ics");
+
+    final InMemoryLdapServer inMemoryLdapServer = new Builder().
+        withLdapPort(LDAP_PORT)
+        .build();
+    inMemoryLdapServer.start();
+    inMemoryLdapServer.initializeBasicData();
```
matteobaglini, line 63: Hi @frisonisland, I don't like calling `initializeBasicData` after `start`.
I propose:
- if `initializeBasicData` is a mandatory step invoke it directly inside start.
- If it is optional let put a new `withInitializeBasicData` method in the Builder and the invoke or not the logic in the `start`.

store/src/main/java/com/zextras/mailbox/LikeXmlJettyServer.java
frisonisland: Right, updated

store/src/test/java/com/zextras/mailbox/SampleLocalMailbox.java
frisonisland: I agree but I have to check other tests since it seems the two methods are public and used in different scenarios

store/src/test/java/com/zextras/mailbox/LikeXmlJettyServerTest.java
frisonisland: Do you mean having different tests with same setup but different assert?
Like:
```suggestion
  @BeforeEach
  void setUp() {
    config = mock(Config.class);
    localServer = mock(com.zimbra.cs.account.Server.class);
    when(config.getHttpNumThreads()).thenReturn(20);
    when(config.getMailboxdSSLProtocols()).thenReturn(new String[] {"TLS1.2"});
    when(config.getSSLExcludeCipherSuites()).thenReturn(new String[] {"^TLS_RSA_.*"});
    when(config.getSSLIncludeCipherSuites()).thenReturn(new String[] {});
    when(config.getHttpConnectorMaxIdleTimeMillis()).thenReturn(3000);
  }

  @Test
  void shouldCreateJettyServer_withUserConnector() throws InstantiationException {
    final Server server = new Builder(config, localServer).build();
    
    final Connector[] connectors = server.getConnectors();
    final Connector userConnector = connectors[0];
    assertEquals("http/1.1", userConnector.getProtocols().get(0));
    assertEquals(3000, userConnector.getIdleTimeout());
  }
```

M0Rf30 (changes requested):

packages/appserver-service/carbonio-appserver.service
```diff
@@ -7,33 +7,17 @@ PartOf=carbonio-appserver.target carbonio-ce.target carbonio.target
...
-  --module-path /opt/zextras/mailboxd/common/endorsed \
-  -Djetty.base=/opt/zextras/mailboxd \
-  -Djetty.home=/opt/zextras/common/jetty_home \
-  -DSTART=/opt/zextras/mailboxd/etc/start.config \
-  -jar /opt/zextras/common/jetty_home/start.jar \
-  --module=zimbra,server,mail,servlet,servlets,jsp,jstl,jmx,resources,websocket,ext,plus,rewrite,continuation,webapp,setuid \
-  jetty.home=/opt/zextras/common/jetty_home \
-  jetty.base=/opt/zextras/mailboxd \
-  /opt/zextras/mailboxd/etc/jetty.xml
+            -Dfile.encoding=UTF-8 \
```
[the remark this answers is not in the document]
frisonisland: I applied an shfmt. However we do not use such usually in our repos and do not really check linting anymore

store/src/test/java/com/zextras/mailbox/LikeXmlJettyServerTest.java
frisonisland: Applied similar changes

M0Rf30 (approved):

PR #501, opened 2024-04-24
feat: [ZIF-1022] Convert the smart-link metrics to send info to posthog (instead of matomo)

Gammax92:
Note: tested also against eu.posthog.com.


frisonisland (approved):

federicorispo:
@Gammax92 @frisonisland Sorry to jump in, but I don't think is a good idea to push the PostHog API key on a public repository https://github.com/zextras/carbonio-mailbox/pull/501/files#diff-c18ef9eb6eb17fef40e4f482bc64ccb37c5df9edb6eef7e24485f9c33238f964R12

I also suggest to change it since it could be already compromised

domenicostragliotto:
> @Gammax92 @frisonisland Sorry to jump in, but I don't think is a good idea to push the PostHog API key on a public repository https://github.com/zextras/carbonio-mailbox/pull/501/files#diff-c18ef9eb6eb17fef40e4f482bc64ccb37c5df9edb6eef7e24485f9c33238f964R12
> 
> I also suggest to change it since it could be already compromised

I was going to write the exact same thing, that's a really big security concern

frisonisland:
> @Gammax92 @frisonisland Sorry to jump in, but I don't think is a good idea to push the PostHog API key on a public repository https://github.com/zextras/carbonio-mailbox/pull/501/files#diff-c18ef9eb6eb17fef40e4f482bc64ccb37c5df9edb6eef7e24485f9c33238f964R12
> 
> I also suggest to change it since it could be already compromised

Hi @federicorispo it is a very valid concern. When we used Matomo anyone could send data to it, given a siteId. Now, in PostHog, we discussed the usage of this API key and we decided it is ok to leave it public, since it is valid only for the public endpoint and for a specific project. Open to any other suggestions

domenicostragliotto:
> > @Gammax92 @frisonisland Sorry to jump in, but I don't think is a good idea to push the PostHog API key on a public repository https://github.com/zextras/carbonio-mailbox/pull/501/files#diff-c18ef9eb6eb17fef40e4f482bc64ccb37c5df9edb6eef7e24485f9c33238f964R12
> > I also suggest to change it since it could be already compromised
> 
> Hi @federicorispo it is a very valid concern. When we used Matomo anyone could send data to it, given a siteId. Now, in PostHog, we discussed the usage of this API key and we decided it is ok to leave it public, since it is valid only for the public endpoint and for a specific project. Open to any other suggestions

So you're still allowing anyone to send data to that endpoint, easily polluting them for that particular project, not even loading it from ENV or configfile

federicorispo:
@frisonisland @domenicostragliotto @Gammax92 If something was done in the past does not mean it was right and/or it should be done in the same way. I suggest a meeting because I have a lot of questions about the choice of this tool, about the implementation and the security concerns around it

AlessioCoser (approved):

frisonisland:
> @frisonisland @domenicostragliotto @Gammax92 If something was done in the past does not mean it was right and/or it should be done in the same way. I suggest a meeting because I have a lot of questions about the choice of this tool, about the implementation and the security concerns around it

I think we are not considering few points in here:
-  this public key identifies a project
- this key is for public usage, not private

If you are concerned about polluting analytics of a public product that can be installed by anyone in the world, yes that is a possibility. This API key doesn't allow reading data or do any other stuff. 
Don't be mislead by the name API key, it is the public api key of the project. 

See also:
- https://posthog.com/questions/is-it-ok-to-expose-the-posthog-project-api-key-to-the-public
- https://posthog.com/docs/api/post-only-endpoints

keshavbhatt (approved):
Changes look good to me, except for the worry of potential pollution of the metrics due to the exposure of the API key used for sending data to the endpoint.

store/src/main/java/com/zextras/mailbox/tracking/PostHogTracking.java
```diff
@@ -0,0 +1,38 @@
...
+import java.io.UnsupportedEncodingException;
+import java.net.URI;
+import org.apache.http.client.methods.HttpPost;
+import org.apache.http.entity.StringEntity;
+import org.apache.http.impl.client.CloseableHttpClient;
+import org.apache.http.impl.client.HttpClientBuilder;
+
+public class PostHogTracking implements Tracking {
+
+    static final String SITE_KEY = "phc_egpFZ14OKByQMK51wCTzYp8tLrg0VA8wa2QDagXCjDG";
```
sonersivri, line 12: @Gammax92 this key can be put in localconfig with default value phc_egpFZ14OKByQMK51wCTzYp8tLrg0VA8wa2QDagXCjDG.

store/src/main/java/com/zimbra/cs/service/mail/MailService.java
```diff
@@ -249,7 +249,7 @@ protected Provisioning getProvisioning() {
   }
 
   protected Tracking getTracking() {
-    return new MatomoTracking("https://analytics.zextras.tools");
+    return new PostHogTracking("https://eu.posthog.com");
```
sonersivri, line 252: @Gammax92 this url can be put in localconfig with default value https://eu.posthog.com.

zextrasGiovanniDefacci (approved):

store/src/main/java/com/zextras/mailbox/tracking/PostHogTracking.java
zextrasGiovanniDefacci: We asked @larcara , he does not want to make this value configurable

store/src/main/java/com/zimbra/cs/service/mail/MailService.java
zextrasGiovanniDefacci: We asked @larcara , he does not want to make this value configurable

PR #517, opened 2024-05-17
feat: [COR-916] Use proxy config to make outbound connections through internet

keshavbhatt:
**What's changed:**
- zimbraHttpProxyURL global, server attribute is single cardinal now
- use proxy config in PostHogTracking
- use proxy config in CallToHome
- refactor/cleanup of HttpProxyUtil
- refactor/cleanup of CallToHome
- remove proxy config from PreviewServlet since it is not required (internal connection)
- move ProxyServlet from zimlet package + minor cleanup

**Related PRs:**
- https://github.com/zextras/carbonio-advanced-lib/pull/110
- https://github.com/zextras/carbonio-advanced-core/pull/115
- https://github.com/zextras/carbonio-teams/pull/61


frisonisland (approved):

sonersivri (approved):

store/src/test/java/com/zextras/mailbox/tracking/PostHogTrackingTest.java
```diff
@@ -21,12 +24,14 @@ class PostHogTrackingTest {
...
+    public void startUp() throws Exception {
+        MailboxTestUtil.setUp();
         postHog = startClientAndServer(POSTHOG_PORT);
     }
 
     @AfterEach
-    public void tearDown() throws IOException {
+    public void tearDown() throws Exception {
+        MailboxTestUtil.tearDown();
         postHog.stop();
```
frisonisland, line 35: I don't think these lines are neeeded. IIRC this is an Integration test for the internal PostHog Library that doesn't need any mailbox. Maybe they were added by mistake while refactoring?

store/src/main/java/com/zimbra/cs/service/servlet/proxy/ProxyServlet.java
frisonisland, line 1: Is this the same ProxyServlet from zimlet or does it include new changes?

store/src/test/java/com/zextras/mailbox/tracking/PostHogTrackingTest.java
keshavbhatt: The reason why we have this now is because PosthogTracking class now uses HttpProxyUtil that uses provisioning.

store/src/main/java/com/zimbra/cs/service/servlet/proxy/ProxyServlet.java
keshavbhatt: Its the same class. We just preserved it.(moved out from zimlet package that is schedule for removal). The logic is almost same just minor refactoring. 
We could change some stuff here later though.

store/src/test/java/com/zextras/mailbox/tracking/PostHogTrackingTest.java
frisonisland: But why? Is it possible to avoid such strong constraint? It's ok to depend on static methods, but I don't think it's a good idea to bring all legacy Zimbra dependencies with it.

store/src/main/java/com/zextras/mailbox/tracking/PostHogTracking.java
```diff
@@ -18,7 +19,10 @@ public PostHogTracking(String endPoint) {
 
     @Override
     public void sendEventIgnoringFailure(Event event) {
-        try (CloseableHttpClient client = HttpClientBuilder.create().build()) {
+        var clientBuilder = HttpClientBuilder.create();
+        HttpProxyUtil.configureProxy(clientBuilder);
+
+        try (CloseableHttpClient client = clientBuilder.build()) {
```
frisonisland, line 25: I suggest using a factory to avoid such strong cohesion, for example by making PostHogTracking depend on a httpClient factory. In this case the real factory could return an httpclient already configured with all our needs.
```suggestion
        try (CloseableHttpClient client = httpClientFactory.create()) {
```

keshavbhatt: We will try this approach. Thanks

sonersivri (approved):

store/src/test/java/com/zextras/mailbox/tracking/PostHogTrackingTest.java
keshavbhatt: Removed coupling.

frisonisland (approved):

aheeva-yuliya (approved):

PR #528, opened 2024-06-20
fix: [CO-1180] Remove TNEF converter to prevent .dat attachment duplication

aheeva-yuliya:
**Issue:**
The issue described in [CO-1180] involves a strange duplication of attachment files with a .dat extension. The expected behavior is to send the file itself rather than its expanded content.

**Fix:**
To resolve this issue, the TNEF (Transport Neutral Encapsulation Format) converter responsible for expanding the content of `application/ms-tnef` attachments has been removed. This change ensures that the file is sent as an attachment without unnecessary content expansion.

**Changes Made:**
Removed the TNEF converter to prevent the expansion of `application/ms-tnef` attachments.

**Testing:**
Verified that attachments are no longer duplicated and are sent correctly as files without expanding their content.
Ensured that the removal of the TNEF converter does not impact other attachment handling functionalities.

[CO-1180]: https://zextras.atlassian.net/browse/CO-1180?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


frisonisland (approved):

keshavbhatt (approved):

store/src/main/java/com/zimbra/cs/mime/MimeVisitor.java
```diff
@@ -40,13 +40,6 @@ public abstract class MimeVisitor {
             } catch (Exception e) {
                 ZimbraLog.misc.error("error loading UUENCODE converter", e);
             }
-
-            try {
-                if (LC.zimbra_converter_enabled_tnef.booleanValue())
```
keshavbhatt, line 45: @aheeva-yuliya We can also get rid of this LC key if it is not relevant anymore.

sonersivri (approved):

PR #536, opened 2024-07-01
chore: align devel with main

matteobaglini:


keshavbhatt:
@matteobaglini its too early for this PR. We have to release hotfix from `main` tomorrow. 
We will need this PR when we will release 24.9.0 in future. After hotfix we can do main -> devel and not the opposite :)

matteobaglini:
> @matteobaglini its too early for this PR. We have to release hotfix from `main` tomorrow. We will need this PR when we will release 24.9.0 in future. After hotfix we can do main -> devel and not the opposite :)

Sorry my mistake, thanks @keshavbhatt

PR #539, opened 2024-07-01
chore: use new and builtin argon2 module included in 2.5.x

M0Rf30:
Seems that adding migrate file is not enough. So I replace the .la module entry directly within ldif file. Let me know if it's a good approach. I don't know ldap technicatlities

Same stuff for carbonio-advanced-lib occurrance
https://github.com/zextras/carbonio-advanced-lib/pull/119


frisonisland (approved):

keshavbhatt (approved):
Seems okay to me.

This page also suggest that the built in module available was renamed to argon2 from pw-argon2. Page also mentions how we can confirm the load of module. See section that mention "how to hash password using `slappasswd` util loading the module manually"

https://wiki.zionetrix.net/informatique:reseau:ldap:pw-argon2

PR #548, opened 2024-07-10
feat: [CO-1343] Update LDAP Module Configuration on upgrade

keshavbhatt:
Ensure LDAP module configuration is updated whenever modifications are applied to the modules loaded by **carbonio-openldap** by executing `ldapmoduleconfupdate` (added in linked PR below) in the post-installation hook of the **carbonio-directory-server** package upon upgrades. 

Also removes `migrate20240201-AddOpenLDAP25.pl` migration script that was added in 4ebb0b19edde9abd08131566e13648e3306f5465

**Linked PRs:**
- https://github.com/zextras/carbonio-core-utils/pull/87
- https://github.com/zextras/carbonio-core/pull/75


keshavbhatt:
@M0Rf30 I would like to know from you the status of existing migration (perl script). IMO we can safely remove it. What you say?

keshavbhatt:
> @M0Rf30 I would like to know from you the status of existing migration (perl script). IMO we can safely remove it. What you say?

Decided to remove migrate20240201-AddOpenLDAP25.pl since it is not required.

aheeva-yuliya (approved):

packages/directory-server/PKGBUILD
```diff
@@ -130,6 +132,8 @@ postinst__rocky_9() {
 
   if [ "$1" -eq 2 ]; then
     echo "* Starting LDAP upgrade..."
+
+    su - zextras -c "/opt/zextras/libexec/ldapmoduleconfupdate"
```
frisonisland, line 136: Why is there no ldap stop on rocky9? But in rocky8 and apt there is
Are these migrations supposed to run before stopping ldap, while it is stopped or after schema update and restart?

keshavbhatt: @frisonisland the stop ldap was already there, I don't know why rockey9 do not have this. Maybe @M0Rf30 can answer that. 

**About the placement of `ldapmoduleconfupdate` :**
- we need to keep it before stopping slapd so that we can make sldap load newly added modules before applying OLC updates in one of the following steps.

M0Rf30: > @keshavbhatt Why is there no ldap stop on rocky9? But in rocky8 and apt there is Are these migrations supposed to run before stopping ldap, while it is stopped or after schema update and restart?

Because of this
https://github.com/zextras/carbonio-mailbox/blob/devel/packages/directory-server/PKGBUILD#L134

That's the same motivation that caused you to use ldap stop (instead of restart)

frisonisland: It's not still clear to me, but is unrelated to the PR, we can check it another time

PR #557, opened 2024-07-23
CO-1294: new SOAP api to autocomplete internal users with a particular feature enabled

mbarto:
The original requirement for this new API is to let user create a WSC chat 1-1 or group, by searching only internal users that have the module enabled.

The implementation is done using the Provisioning SearchDirectory functionality with an ad-hoc filter.

**Endpoint:**
```
https://<host>/service/soap/SearchUsersByFeature
```

**Sample request:**
```xml
<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope">
  <soap:Header>
    <context xmlns="urn:zimbra">
      <authToken>${ZM_AUTH_TOKEN}</authToken>
      <nosession/>
      <userAgent name="zmsoap"/>
      <authTokenControl voidOnExpired="0"/>
    </context>
  </soap:Header>
  <soap:Body>
    <SearchUsersByFeatureRequest name="b" feature="CHATS" xmlns="urn:zimbraAccount"/>
  </soap:Body>
</soap:Envelope>
```

**Sample response:**
```xml
<soap:Envelope xmlns:soap="http://www.w3.org/2003/05/soap-envelope">
    <soap:Header>
        <context xmlns="urn:zimbra"/>
    </soap:Header>
    <soap:Body>
        <SearchUsersByFeatureResponse xmlns="urn:zimbraAccount">
            <account name="baz@demo.zextras.io" id="8f8e1c63-2f55-483d-be36-e587ae3ba8b9">
                <a n="displayName">Surname Firstname</a>
                <a n="uid">baz</a>
                <a n="mail">baz@demo.zextras.io</a>
            </account>
        </SearchUsersByFeatureResponse>
    </soap:Body>
</soap:Envelope>
```

The first implementation:
 * does not implement pagination of results, only a limit parameter that defaults to 10 to limit the max # of returned results.
 * sorts by uid ascending (not configurable)
 * only internal accounts of the domain of the requesting user are returned (no distribution lists or groups)
 * accounts with zimbraHideInGal TRUE are not returned
 * autocomplete search (name parameter) is done in the account uid, displayName and mail fields
 * if the optional feature parameter is specified (currently only CHATS is accepted as a value) only users with the feature enabled (either via COS or account) will be returned (CHATS is mapped to the carbonioFeatureChatsEnabled attribute)
 * the attibutes returned are displayName, mail, uid, in addition to the account zimbraId and main email address

### DIFF
--- a/build_packages.sh
+++ b/build_packages.sh
@@ -22,6 +22,14 @@ then
     -v $(pwd):/tmp/staging \
     docker.io/m0rf30/yap-ubuntu-jammy:1.8 \
     build ubuntu-jammy /tmp/staging/packages
+elif [[ $OS == "ubuntu-focal" ]]
+then
+  docker run -it --rm \
+    --entrypoint=yap \
+    -v $(pwd)/artifacts/ubuntu-focal:/artifacts \
+    -v $(pwd):/tmp/staging \
+    docker.io/m0rf30/yap-ubuntu-focal:1.8 \
+    build ubuntu-focal /tmp/staging/packages
 elif [[ $OS == "rocky-8" ]]
 then
   docker run -it --rm \
--- a/common/src/main/java/com/zimbra/common/localconfig/LC.java
+++ b/common/src/main/java/com/zimbra/common/localconfig/LC.java
@@ -117,6 +117,14 @@ public final class LC {
   public static final KnownKey localized_msgs_directory =
       KnownKey.newKey("${zimbra_home}/conf/msgs");
 
+  @Supported
+  public static final KnownKey localized_client_msgs_directory =
+      KnownKey.newKey("${mailboxd_directory}/webapps/zimbraAdmin/WEB-INF/classes/messages");
+
+  @Supported
+  public static final KnownKey skins_directory =
+      KnownKey.newKey("${mailboxd_directory}/webapps/zimbra/skins");
+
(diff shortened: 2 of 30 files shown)


zextrasGiovanniDefacci (approved):

keshavbhatt (approved):
@mbarto the name of this API sounds little odd, if it possible can we reflect "feature" in the name?

here are few suggestions:
- GetUsersByFeature
- SeachUsersByFeature
- QueryUsersByFeature 

Consider this suggestion non blocking.

mbarto:
> @mbarto the name of this API sounds little odd, if it possible can we reflect "feature" in the name?
> 
> here are few suggestions:
> 
> * GetUsersByFeature
> * SeachUsersByFeature
> * QueryUsersByFeature
> 
> Consider this suggestion non blocking.

@keshavbhatt very good suggestion, I didn't like the original name too. I think I will rename it to SearchUsersByFeature

nomanAli95 (approved):

nomanAli95:
@mbarto could you please update also the description of this PR with the new endpoint, new request name, new response name etc?

mbarto:
> @mbarto could you please update also the description of this PR with the new endpoint, new request name, new response name etc?

done

PR #558, opened 2024-07-29
feat: [CO-1357] emit user status changed event on message broker

galvagnimatteo:
- Publish UserStatusChanged message using message-broker-sdk on account status changed callback
- Add UT

Refs: CO-1357


store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
```diff
@@ -0,0 +1,71 @@
...
+  public ServiceDiscoverHttpClient withToken(String token) {
+    this.token = token;
+    return this;
+  }
+
+  public Try<String> getConfig(String configKey) {
+    try (CloseableHttpClient httpClient = HttpClients.createMinimal()) {
+      HttpGet request = new HttpGet(serviceDiscoverURL + configKey);
+      request.setHeader("X-Consul-Token", token);
+      CloseableHttpResponse response = httpClient.execute(request);
```
keshavbhatt, line 53: Consider closing the response with `try-with-resources` statement.

store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
```diff
@@ -0,0 +1,71 @@
...
+          response.getEntity().getContent(),
+          StandardCharsets.UTF_8
+        );
+
+        String value = new ObjectMapper().readTree(bodyResponse).get(0).get("Value").asText();
+        String valueDecoded = new String(Base64.decodeBase64(value), StandardCharsets.UTF_8).trim();
+
+        return Try.success(valueDecoded);
+      }
+      return Try.failure(new UnAuthorized());
```
keshavbhatt, line 66: We should consider adding suitable log line here to track this failure.

store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
```diff
@@ -0,0 +1,71 @@
...
+    } catch (IOException exception) {
+      return Try.failure(new InternalServerError(exception));
```
keshavbhatt, line 68: We should consider adding suitable log line here to track this failure.

store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
```diff
@@ -0,0 +1,71 @@
+// SPDX-FileCopyrightText: 2022 Zextras <https://www.zextras.com>
+//
+// SPDX-License-Identifier: AGPL-3.0-only
+
+package com.zextras.mailbox.client;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.zextras.carbonio.files.exceptions.InternalServerError;
+import com.zextras.carbonio.files.exceptions.UnAuthorized;
+import com.zimbra.common.util.ZimbraLog;
```
keshavbhatt, line 10:
```suggestion
```
Lets remove it until we use it :)

store/src/main/java/com/zimbra/cs/account/callback/AccountStatus.java
```diff
@@ -55,14 +63,14 @@ public void preModify(CallbackContext context, String attrName, Object value,
...
             return;
         }
 
         if (!context.isCreate()) {
             if (entry instanceof Account) {
                 try {
+                    publishStatusChangedEvent((Account)entry);
                     handleAccountStatusClosed((Account)entry);
                 } catch (ServiceException se) {
                     // all exceptions are already swallowed by LdapProvisioning, just to be safe here.
```
keshavbhatt, line 76: We can remove this comment and update the log line below to be more generic since we have new statement in try block now.

I would avoid logging a generic message though. 
It would be nice if we would catch these exceptions in the methods themselves and add logging there instead.

for example, in `publishStatusChangedEvent` method itself:
```Java
try {
    ...
    ...
    messageBrokerClient.publish(new UserStatusChanged(userId, status.toUpperCase()));
    ZimbraLog.account.info("Published status changed event for user: " + userId);
} catch (Exception e) {
    ZimbraLog.account.error("Failed to publish status changed event for user: " + userId, e);
}
```

keshavbhatt (approved):
LGTM, I have just added a few comments that focuses on:
- improved logging and
- resource management

store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
```diff
@@ -50,21 +53,25 @@ public Try<String> getConfig(String configKey) {
     try (CloseableHttpClient httpClient = HttpClients.createMinimal()) {
       HttpGet request = new HttpGet(serviceDiscoverURL + configKey);
       request.setHeader("X-Consul-Token", token);
-      CloseableHttpResponse response = httpClient.execute(request);
+      try(CloseableHttpResponse response = httpClient.execute(request)) {
```
federicorispo, line 56: The `CloseableHttpClient` is already in a `try-with-resources` statement, and, when the execution reaches the the `finally` block, the following methods are called in sequence:
- `CloseableHttpClient#close()`
- `PoolingHttpClientConnectionManager#close()`
- [`PoolingHttpClientConnectionManager#shutdown()`](https://github.com/apache/httpcomponents-client/blob/4.5.x/httpclient/src/main/java/org/apache/http/impl/conn/PoolingHttpClientConnectionManager.java#L409) -> which closes each connection and then shuts down the pool.

At the moment, the carbonio-mailbox is using the apache httpclient version 4.5.x. Although the close mechanism is better handled in the 5.3.x version, I believe it is unnecessary to add another try-with-resources around the `CloseableHttpResponse`

keshavbhatt: Thanks @federicorispo, I confirm that the resources (socket) is closed as soon as CloseableHttpClient#close() is called. The official documentation still suggests to release the response see https://hc.apache.org/httpcomponents-client-4.5.x/quickstart.html and some example they provide here https://hc.apache.org/httpcomponents-client-4.5.x/examples.html

store/src/test/java/com/zextras/mailbox/callbacks/AccountStatusChangedCallbackTest.java
```diff
@@ -0,0 +1,86 @@
...
+import static org.mockito.ArgumentMatchers.any;
+
+class AccountStatusChangedCallbackTest {
+
+  AccountStatus accountStatus;
+
+  @BeforeEach
+  void setUp() throws Exception {
+    MailboxTestUtil.setUp();
+    accountStatus = new AccountStatus();
```
frisonisland, line 35: What do you think about defining an AccountStatus with the messageBrokerClient and Consul token provider as dependencies? This way we can unit test the component and avoid static mocking.
Unfortunately the empty constructor still needs to stay as this class is instantiated through reflection and will have to pass  FileSystem Consul Provider inside it. That part could be tested in integration, for example by creating an account and checking we receive a call to the broker.  Example:
```
 public AccountStatus() {
   this.consulTokenProvider = new FileSystemConsulTokenProvider();
   this.messageBrokerCLient = new MessageBrokerClient("127.001", ...);
};

 public AccountStatus(ConsulTokenProvider consulTokenProvider, MessageBrokerCient client) {
   this.consulTokenProvider = ...;
   this.messageBrokerCLient = ...;
};
```

store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
federicorispo: @keshavbhatt Good point! However, in the apache example, the httpclient is created outside the `try-with-resources` statement to be reused later. In our case the httpClient is created, used and destroyed (with its pool and connections)

store/src/test/java/com/zextras/mailbox/callbacks/AccountStatusChangedCallbackTest.java
galvagnimatteo: That would be ideal, indeed. However, we don't know how to define those provider such that they are injected inside the constructor. Could you point out how to properly inject them inside AccountStatus? We saw that Files' client is provided directly by MailService#getFilesClient and passed in classes that needs it that are then registered in a dispatcher, but it's not really the case here.

galvagnimatteo: @frisonisland Any update regarding my last comment?

frisonisland: Sorry for the late reply, the public constructor with two arguments would be used just for testing for now.
You would have to instantiate it  like you are doing right now, except in constructor. 
Just a suggestion, we can roll with the current changes and implement/try this path in the future.

frisonisland: Sorry, I think we just need the broker dependency, example:
```
public AccountStatus() {
Path filePath = Paths.get("/etc/carbonio/mailbox/service-discover/token");
        String token;
        try {
            token = Files.readString(filePath);
        } catch (IOException e) {
            throw new RuntimeException("Can't read consul token from file", e);
        }

        ServiceDiscoverHttpClient serviceDiscoverHttpClient =
            ServiceDiscoverHttpClient.defaultURL("carbonio-message-broker")
                .withToken(token);
  
  MessageBrokerClient messageBrokerCient =  MessageBrokerClient.fromConfig(
                    "127.78.0.7",
                    20005,
                    serviceDiscoverHttpClient.getConfig("default/username").get(),
                    serviceDiscoverHttpClient.getConfig("default/password").get()
                )
                .withCurrentService(Service.MAILBOX)
 
    this(messageBrokerClient);
};

 public AccountStatus(MessageBrokerClient client) {
   this.messageBrokerClient = client;
};

```
Anyway it's ok and safer to go with current implementation, these changes would need to be well tested in a real environment as I don't fully know what will happen with a early-initialization of the client and consul access.

galvagnimatteo: Since adding dependency injection on callbacks could be an improvement not only for tests but for the codebase in general, I think it's best to keep this PR as-is without changing its logic for a single test and instead implementing dependency injection in the future with a separate tech improvement and in a more general way (for every callback instead of just this one, and for every callback's test, for example).

frisonisland (approved):

PR #567, opened 2024-08-22
feat: [CO-1425] Improve argon2 and PID file migration for LDAP

keshavbhatt:
**What has changed:**
- Introduce a pre-flight runner that takes care of running miscellaneous LDAP updates like:
    - module config update
    - pid file path update
- removed unused migration scripts
- add `ldap start` in the pre-install phase

**Related PRs:**
- https://github.com/zextras/carbonio-core/pull/80
- https://github.com/zextras/carbonio-core-utils/pull/100


frisonisland (approved):

frisonisland (approved):

packages/directory-server/PKGBUILD
```diff
@@ -167,6 +155,11 @@ postinst__rocky_8() {
   fi
 }
 
+preinst__rocky_9() {
+  if [ "$1" -eq 2 ]; then
+    su - zextras -c "/opt/zextras/bin/ldap start || true"
```
M0Rf30, line 160: systemd unit is enabled by default on both rocky-9 and ubuntu-noble
You need to use
```
  if [ "$1" -eq 2 ]; then
    if [ -d /run/systemd/system ]; then
      systemctl start carbonio-openldap.service >/dev/null 2>&1 || :
    fi
  fi
```

and also:
on upgrade is ldap still running and available? is this really needed?

M0Rf30 (changes requested):

PR #571, opened 2024-08-29
feat: [CO-1430] emit user delete event on message broker when user is deleted

galvagnimatteo:
- Publish UserDeleted event when user is deleted
- Adapt delete user IT

Refs: CO-1430


store/src/main/java/com/zimbra/cs/service/admin/DeleteAccount.java
```diff
@@ -100,4 +110,39 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
+    } catch (IOException e) {
+      throw new RuntimeException("Can't read consul token from file", e);
+    }
+
+    ServiceDiscoverHttpClient serviceDiscoverHttpClient =
+        ServiceDiscoverHttpClient.defaultURL("carbonio-message-broker")
+            .withToken(token);
+
+    try {
+      MessageBrokerClient messageBrokerClient = MessageBrokerClient.fromConfig(
```
frisonisland, line 130: I recommend passing the messageBrokerClient in constructor in this case. Unlike AccontStatus, here it is possible to do it, you can add it in the AdminService class

frisonisland (approved):

zextrasGiovanniDefacci (approved):

keshavbhatt (approved):

PR #577, opened 2024-09-02
fix: versioning

frisonisland:
These changes address three issues:
- alignment of jython-libraries module version
- set of profile -pProd or default (devel) depending if building tag or not
- Fix dependencies version using project.version variable, according to: https://maven.apache.org/maven-ci-friendly.html

We noticed that releasing a non SNAPSHOT version produces a pom with SNAPSHOT dependencies, for example zm-mailbox-24.9.1.pom contains zm-common-24.9.1-SNAPSHOT. 
This causes projects using mailbox as dependency will fail downloading transitive dependencies (zm-store-24.9.1-SNAPSHOT does not exist).
Using ${project.version} seems to be the supported way by Maven.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -9,11 +9,8 @@ def mvnCmd(String cmd) {
     sh 'mvn -B -s settings-jenkins.xml ' + profile + ' ' + cmd
 }
 def isBuildingTag() {
-    def changeSet = currentBuild.changeSets[0]
-    if (changeSet != null) {
-        return changeSet.getItems().any { item ->
-            item.comment.contains('refs/tags/')
-        }
+    if (env.TAG_NAME) {
+        return true
     }
     return false
 }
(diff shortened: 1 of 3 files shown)


aheeva-yuliya (approved):

zextrasGiovanniDefacci (approved):

sonersivri (approved):

Jenkinsfile
```diff
@@ -164,15 +161,17 @@ pipeline {
...
         }
         stage('Publish to maven') {
             when {
-                buildingTag()
+                expression {
+                    return isBuildingTag()
+                }
             }
             steps {
-                mvnCmd('$BUILD_PROPERTIES_PARAMS deploy -DskipTests=true -Pprod')
```
sonersivri, line 175: @frisonisland Is this work?

frisonisland: It shouldn't, because the profile is already set in mvnCmd() function.
mvnCmd() returns:
-  a maven command with -Pprod if building tag
-  a maven command with -Pdev in any other case

This is why I decided to remove -Pprod from here

sonersivri (approved):

PR #599, opened 2024-09-17
fix: [CO-1453] SearchAllDomainsByFeature sometime fails to complete accounts while adding to a group chat on wsc

mbarto:
* introduced global config carbonioSearchAllDomainsByFeature (FALSE/TRUE, defaults to FALSE) to extends search to all domains (defaults to the user domain only)

```
zmprov mcf carbonioSearchAllDomainsByFeature TRUE
```

 * fixed search to include users on default COS, when default COS has WSC enabled


sonersivri:
@mbarto changes should be added to CHANGELOG.md
`release-it --ci` can be used.

matteobaglini (approved):

keshavbhatt:
@mbarto can you also please update the title of the PR to reflect it is related to `SearchAllDomainsByFeature` rather than `AutoComplete` which is a totally different API.

keshavbhatt:
@mbarto have you also explored possibility of taking a request attribute/paramter to differ the API response? I would add an attribute for this only if it is really required.

keshavbhatt (approved):

drazen04 (approved):

mbarto:
> @mbarto have you also explored possibility of taking a request attribute/paramter to differ the API response? I would add an attribute for this only if it is really required.

@keshavbhatt we evaluated this option, but it is a security concern to let user search on all domains if not intended, so we decided this is something the admin needs to configure globally through a specific config

mbarto:
> @mbarto can you also please update the title of the PR to reflect it is related to `SearchAllDomainsByFeature` rather than `AutoComplete` which is a totally different API.

changed, thx

mbarto:
> @mbarto changes should be added to CHANGELOG.md `release-it --ci` can be used.

done

PR #601, opened 2024-09-19
chore: publish sources jars

zextrasGiovanniDefacci:


keshavbhatt (approved):

keshavbhatt:
Hi @zextrasGiovanniDefacci , could you clarify why the source jars are needed in the verify phase? It would be really helpful to have this kind of context included in the PR description for future reference. 

Thanks!

zextrasGiovanniDefacci:
> Hi @zextrasGiovanniDefacci , could you clarify why the source jars are needed in the verify phase? It would be really helpful to have this kind of context included in the PR description for future reference.
> 
> Thanks!

Hi @keshavbhatt, I just followed the example in https://maven.apache.org/plugins/maven-source-plugin/usage.html#installing-the-sources-along-with-your-artifact. I believe the verify phase is required for publishing test jars. Probably we can use a different phase while publishing sources for non-test artifacts, but I'm not an expert on this topic.

sonersivri (approved):

PR #637, opened 2024-12-09
fix: [CO-1720] delete account from mailbox only after files finished deleting blobs from powerstore

galvagnimatteo:
- Added a message broker consumer for a confirmation event emitted by files once blobs are deleted from powerstore
- Updated service discover client to get if a service is installed on current infrastructure
- Now the delete account operation, if files is installed, does not immediatly delete an account: the real deletion will happen after consumer receives the confirmation event

Refs: CO-1720

### DIFF
--- a/store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
+++ b/store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
@@ -82,7 +82,7 @@ public class ServiceDiscoverHttpClient {
           return Try.failure(new InternalServerError(new Exception("Unexpected response status: " + response.getStatusLine().getStatusCode())));
         }
         String bodyResponse = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
-        ZimbraLog.mailbox.info("[DELETE] Service discover response: {}", bodyResponse);
+        ZimbraLog.store.info("[DELETE] Service discover response: {}", bodyResponse);
         if (bodyResponse.equals("[]")) {
           return Try.success(false);
         } else {
--- a/store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
+++ b/store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
@@ -7,6 +7,7 @@ package com.zextras.mailbox.client;
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.zextras.carbonio.files.exceptions.InternalServerError;
 import com.zextras.carbonio.files.exceptions.UnAuthorized;
+import com.zimbra.common.util.ZimbraLog;
 import io.vavr.control.Try;
 import org.apache.commons.codec.binary.Base64;
 import org.apache.commons.io.IOUtils;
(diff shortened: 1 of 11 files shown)


store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
```diff
@@ -7,6 +7,7 @@
 import com.fasterxml.jackson.databind.ObjectMapper;
 import com.zextras.carbonio.files.exceptions.InternalServerError;
 import com.zextras.carbonio.files.exceptions.UnAuthorized;
+import com.zimbra.common.util.ZimbraLog;
```
sonersivri, line 10: remove unused import

store/src/main/java/com/zextras/mailbox/client/ServiceDiscoverHttpClient.java
```diff
@@ -75,4 +70,27 @@ public Try<String> getConfig(String configKey) {
...
+      HttpGet request = new HttpGet(url);
+      request.setHeader("X-Consul-Token", token);
+      try (CloseableHttpResponse response = httpClient.execute(request)) {
+        if (response.getStatusLine().getStatusCode() != HttpStatus.SC_OK) {
+          logger.error("Unexpected response status: {}", response.getStatusLine().getStatusCode());
+          return Try.failure(new InternalServerError(new Exception("Unexpected response status: " + response.getStatusLine().getStatusCode())));
+        }
+        String bodyResponse = IOUtils.toString(response.getEntity().getContent(), StandardCharsets.UTF_8);
+        if (bodyResponse.equals("[]")) {
+          return Try.success(false);
```
sonersivri, line 86:
```java
return Try.success(!"[]".equals(bodyResponse));
```
can be used

store/src/main/java/com/zextras/mailbox/messageBroker/MessageBrokerFactory.java
```diff
@@ -22,15 +22,15 @@ public static MessageBrokerClient getMessageBrokerClientInstance()
...
-					ServiceDiscoverHttpClient.defaultURL("carbonio-message-broker")
+					ServiceDiscoverHttpClient.defaultUrl()
 							.withToken(token);
 
 			return MessageBrokerClient.fromConfig(
 							"127.78.0.7",
 							20005,
-							serviceDiscoverHttpClient.getConfig("default/username")
+							serviceDiscoverHttpClient.getConfig("carbonio-message-broker","default/username")
 									.getOrElse("carbonio-message-broker"),
```
sonersivri, line 32: `carbonio-message-broker`  can be constant

store/src/main/java/com/zimbra/cs/service/admin/DeleteAccount.java
```diff
@@ -25,6 +26,11 @@
 import com.zimbra.soap.admin.message.DeleteAccountResponse;
 
 import io.vavr.control.Try;
+
+import java.io.IOException;
```
sonersivri, line 30: remove unused import

store/src/test/java/com/zimbra/cs/account/ProvUtilTest.java
```diff
@@ -208,6 +203,9 @@ void modifyAccount() throws Exception {
     Assertions.assertEquals(expectedOutput, getOperationResult.trim());
   }
 
+  // Static mock of Files doesn't work here, it should probably go inside ProvUtils since runCommand calls its main method,
+  // but that is not a test class so for now I just disable the test.
+  @Disabled("This test is failing because of the static call to Files.exists() in the deleteAccount method")
   @Test
```
sonersivri, line 209: Do not disable test

keshavbhatt (approved):

store/src/test/java/com/zimbra/cs/account/ProvUtilTest.java
galvagnimatteo: As written in the above comment this test cannot be enabled since it always breaks. This is because it needs the static mock of the Files class to work, since that class is used to retrieve the token to be used in ServiceDiscoverHttpClient. I did not find any way to mock that class in the correct context.

PR #643, opened 2024-12-13
fix: retry consumer init if failed and handle message broker factory exceptions

galvagnimatteo:
- MessageBrokerClientFactory now returns a MessageBrokerClient that is working and healthy (added generic exception handler to catch runtimes and add healthcheck call)
- DeletedUserFilesConsumer init will now be retried if it fails to handle message broker downtime on boot
- MessageBrokerClient is now created for each publish event request to handle a permanently broker client if initialized and passed on boot while message broker was down

Refs: CO-1757


federicorispo (approved):
:rocket:

matteobaglini (approved):

PR #657, opened 2025-01-23
refactor: [CO-1767] extract web.xml

frisonisland:
# Goal
web.xml is a templated file, containing LDAP attributes, managed by zmconfigd.
Since the mailbox already depends on LDAP, this instrumentation has been moved to code in order to simplify the management of the application.

### Autodiscover
There won't be any effect to the autodiscover endpoint.
The Autodiscover endpoint is managed by the rewrite rule in MailboxServer.java class. This rule redirects autodiscover endpoints (e.g.: /autodiscover/*) to /service/extension/autodiscover which is managed by extensions.

### Bonus
- Able to start the whole mailbox locally with "little" instrumentation (see MailboxServerTest)

### Other
With these changes API declaration will move from configuration based (xml), to programmatic, making possible in future to not rely on reflection, but having things checked at build time. An example: The admin API layer could be declared as new AdminSoap(...).

### TODO
- [ ] remove web.xml references and files once everything is well tested, also from here https://github.com/zextras/carbonio-core-utils/blob/devel/conf/configd.cf


sonersivri:
@frisonisland Have you ever tested advanced api endpoints and activesync?

keshavbhatt (approved):

frisonisland:
> @frisonisland Have you ever tested advanced api endpoints and activesync?

Hi @sonersivri not yet, we will before merging these changes. We found out the ActiveSync is managed by AutoDiscoverServlet and we have already a rule that redirects request coming at /autodiscover/* to /extension/autodiscover (Advanced), so we expect no regression at all

sonersivri:
> > @frisonisland Have you ever tested advanced api endpoints and activesync?
> 
> Hi @sonersivri not yet, we will before merging these changes. We found out the ActiveSync is managed by AutoDiscoverServlet and we have already a rule that redirects request coming at /autodiscover/* to /extension/autodiscover (Advanced), so we expect no regression at all

@frisonisland We can test it together if you want

sonersivri (approved):

PR #664, opened 2025-01-29
feat: remove event driven account deletion from mailbox and replace it with http call

galvagnimatteo:
- Removed event-driven logic for account deletion (deletion event publishing and confirmation event consuming)
- Added HTTP call to Files' deleteAllBlobsAndNodes to replace the event-driven account deletion procedure
- Updated and added relevant tests

Refs: CO-1774


store/src/test/java/com/zimbra/cs/service/admin/DeleteAccountTest.java
```diff
@@ -244,30 +234,36 @@ private void doDeleteAccount(DeleteAccount deleteAccount, Account caller,
...
-		}
+    final String toDeleteId = toDelete.getId();
+    this.doDeleteAccount(deleteAccount, caller, toDeleteId);
+    assertNull(provisioning.getAccountById(toDeleteId));
+	}
+
+  @ParameterizedTest(name = "{0}")
+	@MethodSource("getHappyPathCases")
+	void shouldDeleteUserWhenFilesReturnsError(String testCaseName, Account caller, Account toDelete) throws Exception {
+    Mockito.when(filesClientMock.deleteAllNodesAndBlobs(any(), any())).thenReturn(Try.success(false));
```
frisonisland, line 261: @galvagnimatteo could you add few tests for when filesClient throws an exception?
Example: 
- invoking deleteAllBlobs throws
- invoking deleteAllBlobs does not directly throw but the result is an exception/failure

galvagnimatteo: @frisonisland added the requested tests

frisonisland (approved):

keshavbhatt (approved):

store/src/main/java/com/zimbra/cs/service/admin/DeleteAccount.java
```diff
@@ -83,68 +81,58 @@ public Element handle(Element request, Map<String, Object> context) throws Servi
...
-			throw ServiceException.FAILURE("Delete account " + account.getMail() + " has an error: Unable to check if files is installed", e);
-		}
-
-    if (isFilesInstalled) {
-      // Since Files is installed, this will throw an exception if it fails to publish the event, both if message broker
-      // is down or if it is not installed, to avoid deleting the account without deleting the files.
-      boolean success = publishDeleteUserRequestedEvent(account);
-      if (!success) {
-        throw ServiceException.FAILURE("Delete account " + account.getMail() + " has an error: Failed to publish delete user requested event", null);
+    try {
```
keshavbhatt, line 84: It would be nice to extract this logic into a usecase similar to `DeleteUserUseCase`.

galvagnimatteo: @keshavbhatt Is this necessary considering that this solution will likely change soon according to the edit in CO-1771?

matteobaglini: @galvagnimatteo @keshavbhatt maybe to start, just an extract method "deleteFilesNodesAndBlob" (lines 84-108) could be enough. What do you think?

galvagnimatteo: @matteobaglini works for me; just pushed the changes

PR #682, opened 2025-03-13
use a single MessageBrokerClient instance

zextrasGiovanniDefacci:


store/src/main/java/com/zextras/mailbox/messageBroker/MessageBrokerFactory.java
```diff
@@ -7,46 +7,49 @@
...
-		} catch (Exception e) {
-			throw new CreateMessageBrokerException(e);
-		}
-
-		if (!client.healthCheck()) throw new CreateMessageBrokerException("Message broker healthcheck failed");
-		return client;
-	}
+  private static final String SERVICE_NAME = "carbonio-message-broker";
+
+  private static final Map<Boolean, MessageBrokerClient> clientMap = new ConcurrentHashMap<>();
```
sonersivri, line 19: @zextrasGiovanniDefacci could you check the singleton in MessageBrokerFactoryTest?

store/src/main/java/com/zextras/mailbox/messageBroker/MessageBrokerFactory.java
```diff
@@ -7,46 +7,49 @@
...
+
+  private MessageBrokerFactory() {
+  }
+
+  public static MessageBrokerClient getMessageBrokerClientInstance() {
+    return clientMap.computeIfAbsent(true, _key -> createMessageBrokerClientInstance());
```
sonersivri, line 25: @zextrasGiovanniDefacci Instead of using concurrent hash map think about classic way
```java
private static MessageBrokerClient instance;

public static MessageBrokerClient getMessageBrokerClientInstance() {
    if (instance != null) {
        return instance;
    }
    synchronized (MessageBrokerFactory.class) {
        if (instance == null) {
            instance = createMessageBrokerClientInstance();
        }
     }
     return instance;
} 
```

zextrasGiovanniDefacci: The issue is that this is eagerly evaluated, I'd prefer a lazy version

store/src/main/java/com/zextras/mailbox/messageBroker/MessageBrokerFactory.java
```diff
@@ -7,46 +7,56 @@
...
+    }
+    return instance;
+  }
+
+  // Returns a working and healthy MessageBrokerClient instance or throws an exception
+  private static MessageBrokerClient createMessageBrokerClientInstance() {
+    try {
+      String token = getToken();
+      var client = createMessageBrokerClientInstance(token);
+      if (!client.healthCheck()) throw new CreateMessageBrokerException("Message broker healthcheck failed");
```
sonersivri, line 39: @zextrasGiovanniDefacci curly brace is needed even if it one line

sonersivri (approved):

PR #683, opened 2025-03-20
CO-1921-name-of-attachments-with-cyrillic-and-non-latin-symbols-are-spoiled-when-saved-to-carbonio-files

zextrasGiovanniDefacci:


aheeva-yuliya (approved):

store/src/main/java/com/zimbra/cs/service/mail/FilesCopyHandlerImpl.java
```diff
@@ -164,11 +170,40 @@ private Try<String> getDestinationFolderId(CopyToFilesRequest request) {
...
   private Try<String> getAttachmentName(MimePart attachment) {
-    return Try.of(attachment::getFileName)
+    return Try.of(() -> getRawFileName(attachment))
         .onFailure(ex -> mLog.error(ex.getMessage()))
         .recoverWith(ex -> Try.failure(ServiceException.FAILURE("Cannot get file name.", ex)));
   }
 
+  private static String getRawFileName(MimePart part) throws MessagingException {
+    String filename = null;
+    String s = part.getHeader("Content-Disposition", (String)null);
```
keshavbhatt, line 180: Is this also handling extended filename(non-ASCII filenames using filename* with UTF-8 and percent encoding) defined by RFC 5987 and RFC 6266 ?

```
 Content-Disposition: attachment;
                          filename="EURO rates";
                          filename*=utf-8''%e2%82%ac%20rates
  ```

**Review by sonersivri — APPROVED** [review]

File: store/src/main/java/com/zimbra/cs/service/mail/FilesCopyHandlerImpl.java (hunk 1 above)
Comment (aheeva-yuliya) [reply store/src/main/java/com/zimbra/cs/service/mail/FilesCopyHandlerImpl.java]: Hi @keshavbhatt,

As per the current implementation, the `ContentDisposition` class **does not natively handle** the filename* parameter defined by RFC 5987, which is used for extended non-ASCII filenames with UTF-8 and percent encoding.

I started using the `ContentDispositionParser` class to retrieve the filename instead.

I haven't written any specific tests for extended filenames in the `getRawFileNam`e method, as this functionality is already covered in the `ContentDispositionParserTest.`

Comment (keshavbhatt) [reply store/src/main/java/com/zimbra/cs/service/mail/FilesCopyHandlerImpl.java]: > I haven't written any specific tests for extended filenames in the getRawFileName method, as this functionality is already covered in the ContentDispositionParserTest.

Hi @aheeva-yuliya 

That makes sense. Since `ContentDispositionParserTest` already covers extended filenames, additional tests in getRawFileName might not be necessary. However, adding a test here could serve as documentation to confirm that `FilesCopyHandlerImpl` correctly handles extended filenames. I saw that we already added negative tests about broken encoding, I propose the following positive test as well.

```Java
  @Test
  void shouldReturnFileNameForValidContentDispositionEncoding()
      throws MessagingException {
    String contentDispositionHeader = "attachment; filename*=UTF-8''%E6%96%87%E4%BB%B6.txt";
    when(mimePart.getHeader("Content-Disposition", null)).thenReturn(contentDispositionHeader);

    String filename = FilesCopyHandlerImpl.getRawFileName(mimePart);

    assertEquals("文件.txt", filename);
  }
```

keshavbhatt (approved):

PR #689, opened 2025-04-02
chore: standalone mailbox docker

frisonisland:
Allows starting a mailbox for development along with (bare minimum configuration):
- OpenLDAP
- Postfix
- Nginx and web-ui packages
- MariaDB

See README.md for further details.

Some files have potential to be reused, such as the database creation (.sql) and zmldapschema, which originally comes from here: https://github.com/zextras/carbonio-core-utils/blob/devel/src/libexec/ldapschema.sh


docker/standalone/mailbox/entrypoint.sh
```diff
@@ -0,0 +1,17 @@
...
+#
+
+sed -i -e "s#LDAP_URL#${LDAP_URL}#g" /localconfig/localconfig.xml
+sed -i -e "s/LDAP_ROOT_PASSWORD/${LDAP_ROOT_PASSWORD}/g" /localconfig/localconfig.xml
+sed -i -e "s/LDAP_ADMIN_PASSWORD/${LDAP_ADMIN_PASSWORD}/g" /localconfig/localconfig.xml
+sed -i -e "s/MARIADB_ROOT_PASSWORD/${MARIADB_ROOT_PASSWORD}/g" /localconfig/localconfig.xml
+sed -i -e "s/MARIADB_URL/${MARIADB_URL}/g" /localconfig/localconfig.xml
+sed -i -e "s/MARIADB_PORT/${MARIADB_PORT}/g" /localconfig/localconfig.xml
+
+cat /localconfig/localconfig.xml
```
matteobaglini, line 15: Hi @frisonisland, is `cat` call just for debug reason? Can we remove it?

frisonisland: Yes, thanks

PR #690, opened 2025-04-03
chore: [IN-868] pom.xml: update com.fasterxml.jackson to 2.12.7.2

M0Rf30:
Refs: IN-868


frisonisland (approved):

M0Rf30:
In previous attempts with jackson libraries updates, tests always fail in particular jaxbtojson tests.
Looking at original sources, they bumped all the jackson libraries except for jackson-module-jaxb-annotations.

Both build and tests seems to work using updated jackson libraries and leaving only jackson-module-jaxb-annotations to 2.8.11

keshavbhatt (approved):

PR #723, opened 2025-04-24
chore: [CO-2051] remove robots servlet

frisonisland:
See also: https://github.com/zextras/carbonio-proxy/pull/141


sonersivri (approved):

store/src/main/resources/conf/attrs/attrs.xml
```diff
@@ -5490,7 +5490,8 @@ TODO: delete them permanently from here
   </desc>
 </attr>
 
-<attr id="1161" name="zimbraMailKeepOutWebCrawlers" type="boolean" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="7.0.1">
+<attr id="1161" name="zimbraMailKeepOutWebCrawlers" deprecatedSince="25.6.0" type="boolean" cardinality="single" optionalIn="globalConfig,server" flags="serverInherited" since="7.0.1">
```
matteobaglini, line 5493: Hi, why not remove even the attribute (and related code)?

keshavbhatt: Hi @matteobaglini The reason: If someone in existing installs will be using this attribute and we remove it from here the generated schema will not have the attribute definition. LDAP in this case will make the entry using this attribute **read only** unless the schema definition is provided.

PR #770, opened 2025-07-08
chore: [CO-2205] remove zimlet apis

frisonisland:
There are still some usages of Zimlet classes, such as ZimletUtil.
The ZimletUtil is used in the GetInfo API to retrieve all zimlets if the attribute "zimlets" is requested.
A second cleanup/PR will remove these other usages.
In this PR only Zimlet Admin APIs will be removed and the ability to deploy/undeploy zimlets (which is already not supported btw)

After merge:
- [ ] Update mailbox sdk and below projects to use latest sdk
- [ ] https://github.com/zextras/carbonio-user-management
- [ ] https://github.com/zextras/carbonio-storages-service
- [ ] https://github.com/zextras/carbonio-auth-service
- [ ] https://github.com/zextras/carbonio-drive-migration


store/src/main/java/com/zimbra/cs/service/admin/AdminService.java
```diff
@@ -173,17 +173,6 @@ public void registerHandlers(DocumentDispatcher dispatcher) throws ServiceExcept
         AdminConstants.RECALCULATE_MAILBOX_COUNTS_REQUEST, new RecalculateMailboxCounts());
 
     // zimlet
```
matteobaglini, line 175: Hi @frisonisland, very small thing but anyway is this comment still relevant?

frisonisland: I don't think so, thanks :smiley:

matteobaglini (approved):

PR #780, opened 2025-07-21
Chore/IN-881-systemd-restart-services

gjed:
Carbonio.target should restart the services not covered by zmcontrol too

Refs: IN-881

### DIFF
--- a/packages/appserver-service/carbonio-mailbox-admin-sidecar.service
+++ b/packages/appserver-service/carbonio-mailbox-admin-sidecar.service
@@ -11,7 +11,6 @@ ExecStart=/usr/bin/consul connect envoy \
     -token-file /etc/carbonio/mailbox-admin/service-discover/token \
     -admin-bind localhost:0 \
     -sidecar-for carbonio-mailbox-admin
-ExecReload=/usr/bin/kill -HUP $MAINPID
 Restart=on-failure
 RestartSec=15
 ExecReload=/usr/bin/kill -HUP $MAINPID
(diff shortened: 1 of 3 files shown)


packages/directory-server/PKGBUILD
```diff
@@ -1,5 +1,5 @@
 pkgname="carbonio-directory-server"
-pkgver="4.22.0"
+pkgver="4.22.1"
```
frisonisland, line 2: Hello @gjed there is no need to bump the package version at this moment.
We bump all PKGBUILD files simultaneously during RC (release-candidate) by using the automation tool https://github.com/release-it/release-it.
So at RC the package version changes depending on the type of commit we did (feat, fix, etc.).
All packages defined in mailbox end up having the same version (it's also the tag in the repository)
Sorry if it was not clear from the project description! :sweat_smile: 

Dm me if you need further information!

frisonisland (changes requested):

packages/directory-server/PKGBUILD
gjed: Hi @frisonisland, 
alright, I did not understand that the package versions were managed with release-it: I reverted the changes. 

Thanks for the heads up :)

keshavbhatt (approved):

M0Rf30 (approved):

M0Rf30 (changes requested):

M0Rf30 (approved):

gjed:
Some changes introduced are not compatible with older systems.
Centos8 and ubuntu22 still use init.d, therefore some units were not activated at startup.

Solution: revert the `WantedBy=` to `multi-user.target` where modified.

Refs: IN-919 IN-881

keshavbhatt (approved):

M0Rf30 (approved):

PR #807, opened 2025-09-12
refactor: [CO-2522] move RightManager and attrs.xml files (+ ldap schema)

frisonisland:
- moved ldap schema to mailbox-ldap-lib
- moved attrs.xml to attribute-manager
- added rights files/classes generator to attribute-manager (3 NEW classes used just to generate code)
- added right-manager module + classes generation
- moved code generation methods from RightManager to generator classes
- deprecated AttributeManager and RightManager usage with file (loads from resource by default) + deprecated LC. attrs_directory and rights_directory
- **split TargetType to TargetTypeLookup** for LDAP specific operations
- **moved soap TargetType class to SoapTargetType** to avoid cyclic dependencies
- adjusted PKGBUILD
- redefined maven/ant code generation phases (avoid ant when possible)
- **removed ant-build-common and reduced ant commands usage as much as possible**
- declare buildInfo.properties using maven properties and "parsedVersion"

In **bold** the changes that produced most of the noise in this PR.

Notice:
- compared current rights-domainadmin.xml to generated one -> identical

Affects: https://github.com/zextras/zal/pull/56


attribute-manager/test.sh
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+mvn exec:java -Dexec.mainClass="com.zimbra.cs.account.AttributeManagerUtil" -Dexec.args="-a generateGetters -c server -r ./a/ZAttrServer.java"
```
matteobaglini, line 3: Hi @frisonisland, can we make it a real test by adding an assertion? Can be something simple like: no errors printed in stderr and a file exists. Otherwise, we can use the idea of snapshot testing to verify the generated code. Thanks.

PR #943, opened 2026-03-04
feat: [CO-3298] add support for internal REST apis

frisonisland:
### General idea

There are scenarios where a service may need user or mailbox info, but the caller cannot provide a token (service-to-service).

Since Mailbox APIs all need a token, we decided to develop some internal APIs to allow these scenarios.

#### Note on statefullness
Sometimes the requested information is local to each mailbox. 
To avoid having the caller know the exact mailbox where the user lives, and leveraging consul, the internal APIs can:
- route the request to right mailbox
- generate an admin token to make public calls if needed

avoiding any complexity on the caller.

---
### Current purpose

We need to retrieve the mail usage of a user (local information), but we have no token and don't know where the user lives.

---

TODO:
- [x] expose internal endpoint through Consul
- [x] bind internal endpoint to 0.0.0.0 in containers (other containers need to access it)
- [x] verify health endpoint available at internal port
- [x] verify internal endpoint not reachable from admin and user ports

---

## Other changes (tests)

- define local config to avoid starting timers in tests (when mailbox starts it allocates a global timer. on shutodown timer is cancelled. On restart timer cannot restart because it is a static global instance and tests run in same JVM)
- expose method to clear ldap client cache between test runs -> avoids issue of client caching data and throwing "domain already exists", even if the domain does not exist in LDAP

### Related PRs
- https://github.com/zextras/carbonio-storages-service/pull/94


store/src/main/java/com/zextras/mailbox/server/MailboxServerBuilder.java
```diff
@@ -80,15 +81,19 @@ public MailboxServer create() throws InstantiationException {
 			server.addConnector(userHttpConnector);
 
 			final ServerConnector userHttpsConnector = createUserHttpsConnector(server);
-			userHttpsConnector.setName("userHttpsConnector");
+			userHttpsConnector.setName("userHttpsConnector"); // 8080
```
frisonisland: remove comment

store/src/main/java/com/zextras/mailbox/server/MailboxServerBuilder.java
```diff
@@ -80,15 +81,19 @@ public MailboxServer create() throws InstantiationException {
...
 			server.addConnector(userHttpsConnector);
 
 			final ServerConnector adminHttpsConnector = createAdminHttpsConnector(server);
 			server.addConnector(adminHttpsConnector);
 
-			server.addConnector(createMtaAdminHttpsConnector(server));
-			server.addConnector(createExtensionsHttpsConnector(server));
+			server.addConnector(createMtaAdminHttpsConnector(server)); // 7071
```
frisonisland: remove comments

store/src/test/java/com/zextras/mailbox/server/MailboxServerAPITest.java
```diff
@@ -6,105 +6,129 @@
...
 				.call();
 
 		Assertions.assertEquals(200, soapResponse.statusCode());
 	}
 
-}
\ No newline at end of file
+	@Test
+	void internalApiPingShouldNotBeReachableOnUserHttpPort() throws Exception {
+		try (TestHttpClient client = new TestHttpClient()) {
```
frisonisland: use the client extension, not the TestHttpClient with try resources

drazen04 (approved):

keshavbhatt (approved):

gjed (changes requested):
Please also address Copilot comments, especially ones about cut&paste e spdx headers

gjed (approved):

PR #970, opened 2026-03-27
feat: [CO-3400] mailbox check quota

frisonisland:
Adds a class to check quota (centralized).
Any other library is able to override it and implement its own check quota mechanism version.

The old quota check mechanism has been moved to DefaultQuotaCheck implementation


store/src/main/java/com/zimbra/cs/mailbox/Conversation.java
```diff
@@ -856,6 +856,8 @@ static Metadata encodeMetadata(Metadata meta, Color color, int metaVersion, int
     @Override
     protected boolean isQuotaCheckRequired() throws ServiceException {
         Account account = getMailbox().getAccount();
+        // TRUE: allow receiving mails but not sending
```
frisonisland: remove

drazen04 (approved):

PR #978, opened 2026-04-08
chore(ci): bump jenkins-lib-common to 1.5.0

gjed:
## Summary

- Bumps `jenkins-lib-common` library version from `1.4.0` to `1.5.0`
- Adds `buildFlags: ' -sd '` to the `buildStage()` call in the _Build deb/rpm_ stage

## Why `-d` is needed now

`yap` recently started resolving `makedepends` from the `PKGBUILD` spec during the build stage. However, `makedepends` are build-time dependencies of the **package being built**, not of the build environment itself — they are only needed at runtime by the installed package. Checking them during CI builds is unnecessary and can cause failures if those packages aren't present in the build agent.

The `-d` / `--nomakedeps` flag restores the previous behaviour by skipping all `makedepends` checks.

## Flags added

| Flag | Long form | Effect |
|------|-----------|--------|
| `-s` | `--skip-sync` | Skip sync with remotes for package managers |
| `-d` | `--nomakedeps` | Skip all `makedepends` checks |

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -1,5 +1,5 @@
 library(
-        identifier: 'jenkins-lib-common@1.4.0',
+        identifier: 'jenkins-lib-common@1.5.0',
         retriever: modernSCM([
                 $class: 'GitSCMSource',
                 credentialsId: 'jenkins-integration-with-github-account',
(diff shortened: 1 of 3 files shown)


M0Rf30 (approved):

frisonisland:
@gjed I have another PR taking care of tests and it should make them more stable + faster. If packages build in this branch passes let's merge it

frisonisland (approved):

PR #981, opened 2026-04-09
feat: improve internal rest endpoints

frisonisland:
Introduces new internal apis to be used by user-management and storages.

We discussed about the fact that these APIs are returning too much data and we should split the myself or get account in:

- get account basic info (id, email, cos, domain)
- get account preferences
- token validation/authorization/permission

We should also invest some time in finding out if it is possible to have each microservice validate the token (token validity).
However:
- authorization is still performed by mailbox (can a user access another account?)
- admins can manually expire a token. In this case the token sent by a user could be invalid, but this information is available in mailbox/LDAP


store/src/main/java/com/zextras/mailbox/api/rest/resource/AccountResource.java
```diff
@@ -8,56 +8,288 @@
...
+					.filter(entry -> {
+						String key = entry.getKey();
+						return key.startsWith("carbonioWsc") || key.startsWith("carbonioFiles")
+								|| key.startsWith("carbonioTasks") || key.startsWith("carbonioDocs")
+								|| key.startsWith("carbonioPreview");
+					})
+					.collect(Collectors.toMap(Entry::getKey, entry -> entry.getValue().toString()));
+			boolean isExternal;
+			try {
+				isExternal = account.isAccountExternal();
```
frisonisland: Consider to return 500 and log an error instead of relying on default

drazen04 (approved):

PR #989, opened 2026-04-17
chore: [CO-3523] define service name and version for otel collection

frisonisland:
Endpoint is up to the user, so is agent activation (through env variable).
In container environment we set it using JAVA_TOOL_OPTIONS.
In a VM it is done with https://github.com/open-telemetry/opentelemetry-injector if I'm not wrong

Decide:
- [ ] otel.properties path
- [ ] other conventions?

### DIFF
--- a/packages/appserver-service/carbonio-appserver.service
+++ b/packages/appserver-service/carbonio-appserver.service
@@ -8,6 +8,7 @@ PartOf=carbonio-appserver.target
 User=zextras
 Type=notify
 NotifyAccess=main
+Environment=OTEL_JAVAAGENT_CONFIGURATION_FILE=/etc/carbonio/mailbox/otel.properties
 EnvironmentFile=-/opt/zextras/data/systemd.env
 ExecStartPre=echo REWRITE \
   sasl \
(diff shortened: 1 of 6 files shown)


gjed (approved):

packages/appserver-service/carbonio-appserver.service
```diff
@@ -29,6 +29,7 @@ ExecStart=/opt/zextras/common/bin/java \
   -Djava.io.tmpdir=/opt/zextras/mailboxd/work \
   -Djava.library.path=/opt/zextras/lib \
   -Dzimbra.config=/opt/zextras/conf/localconfig.xml \
+  -Dotel.javaagent.configuration-file=/etc/carbonio/mailbox/otel.properties \
   -cp /opt/zextras/mailbox/jars/mailbox.jar:/opt/zextras/mailbox/jars/* \
```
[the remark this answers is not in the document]
frisonisland: outdated

packages/appserver-service/carbonio-appserver.service
```diff
@@ -8,6 +8,7 @@ PartOf=carbonio-appserver.target
 User=zextras
 Type=notify
 NotifyAccess=main
+Environment=OTEL_JAVAAGENT_CONFIGURATION_FILE=/etc/carbonio/mailbox/otel.properties
```
M0Rf30, line 11: I  would avoid this. a sane default value in the code should be sufficient.

M0Rf30 (changes requested):

M0Rf30 (approved):

packages/appserver-service/carbonio-appserver.service
frisonisland: Discussed it's not possible to set it programmatically as the agent needs the env to be defined pre-main startup, I will proceed with merge. Let's monitor how it behaves

PR #1002, opened 2026-04-30
feat: add schema-declared defaults for carbonio*/zimbra* attributes i…

keshavbhatt:
…n Entry fallback


keshavbhatt:
We decided to not ship this.

PR #1014, opened 2026-05-08
fix: reject deregistered auth tokens on internal API 

galvagnimatteo:
The get myself endpoint doesn't check for invalidated token. This PR fixes that.


galvagnimatteo:
Closing: need to target main instead of devel. Reopening as new PR from main.

PR #1015, opened 2026-05-08
chore: fix local build_packages.sh by using release of ubuntu jammy

frisonisland:
- the package build now needs to install packages, but this is not possible without defining a zextras repo. This means that for devel build we should technically use devel, however it requires access to jfrog with credentials, so for now I used release.

We should think if we want to actually mantain these build packages. The downside is that we have to strictly rely to CI if we want to get deb packages and try them in a VM

- [ ] decide if it's worth having these scripts

### DIFF
--- a/build_packages.sh
+++ b/build_packages.sh
@@ -4,13 +4,23 @@
 #
 # SPDX-License-Identifier: GPL-2.0-only
 #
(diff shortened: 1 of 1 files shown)


keshavbhatt (approved):

M0Rf30:
superseeded

PR #1029, opened 2026-05-19
refactor [CO-3718] extract zattr classes from store

frisonisland:
See: https://github.com/zextras/carbonio-directory-server/pull/38

TODO:
- [ ] evaluate impact on ZAL -> needs few changes https://github.com/zextras/zal/pull/108
- [ ] advanced compiles -> compiles with ZAL changes


common/src/main/java/com/zimbra/common/service/ServiceException.java
```diff
@@ -20,7 +21,7 @@
 import org.apache.commons.codec.binary.Hex;
 
 @SuppressWarnings("serial")
-public class ServiceException extends Exception {
+public class ServiceException extends IServiceException {
```
frisonisland, line 24: Perhaps there is no need of an IServiceException. We could just make ZAttrAccount diamond operator use something that extends Exception

store/src/main/java/com/zextras/mailbox/store/ephemeral/RedisEphemeralStore.java
```diff
@@ -46,7 +46,7 @@ public EphemeralResult get(EphemeralKey ephemeralKey, EphemeralLocation location
     final String key = getAccessKey(location, ephemeralKey);
     try (Jedis jedis = jedisPool.getResource()) {
       final String gotResult = jedis.get(key);
-      return new EphemeralResult(ephemeralKey, gotResult);
+      return new EphemeralResult(ephemeralKey, gotResult, ZimbraLog.ephemeral::warn);
```
frisonisland, line 49: This is a side effect of logging. ZimbraLog is not on directory-server, so to preserve logs we need to provide the logger somehow.
Perhaps the same can be achieved with a facade or something like Logger.getLog('ephemeral').warn() of log4j or sfl4j

zextrasGiovanniDefacci (approved):

PR #1036, opened 2026-05-22
 refactor!: collapse Z*/UBID* LDAP class hierarchy

frisonisland:
Mailbox calls Ldap#getInstance and singletons that are problematic because we cannot reliably control the lifecycle.

I found that ZLdap are abstract classes with only one extension (UBID), but ZLdap casts to UBID classes sometimes and this means the abstraction is useless.

Moreover, without this abstraction layer, classes can be directly instantiated, without using `LdapClient#getInstance#getSearchContext`. Just use `new UBIDLDAPContext()`

---
## Changes

Coallapse UBID to ZLdap classes

## Next steps

Proceed removing all side-effect caused by static singletons. Check for example

```private void doRenameDomain(String[] args) throws ServiceException {

    var prov = provUtil.getProvisioning();
    // bug 56768
    // if we are not already using master only, force it to use master.
    // Note: after rename domain, the zmprov instance will stay in "master only" mode.
    if (!provUtil.getUseLdapMaster()) {
      ((LdapProv) prov).alwaysUseMaster();
    }
```
the bug exists because of the architectural decision to keep everything static.

 BREAKING CHANGE: removed public classes from
  com.zimbra.cs.ldap.unboundid: UBIDLdapClient, UBIDLdapContext,
  UBIDAttributes, UBIDMutableEntry, UBIDLdapFilter, UBIDLdapSchema,
  UBIDLdapException, UBIDModificationList, UBIDSearchControls,
  UBIDSearchResultEntry, UBIDSearchResultEnumeration, UBIDSearchScope,
  UBIDLdapFilterFactory. Use the Z* counterparts in
  com.zimbra.cs.ldap. Some package-private members on remaining
  unboundid helpers are now public.


frisonisland:
@zextrasGiovanniDefacci @aheeva-yuliya @sonersivri this has a small impact on zal. If it gets released we must also release zal. How should we proceed?

keshavbhatt (approved):

PR #1037, opened 2026-05-22
ci: immediate bump + cleanup maven deploy

frisonisland:
- moves semantic release as first stage (no need to wait 15 minutes to get tag)
- removes -Pdev (does not exist)
- removes maven snapshot publish
- uses profile to alter changelist, no need to pass it during deploy (see maven pom.xml)

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -13,8 +13,6 @@ boolean isBuildingTag() {
     return env.TAG_NAME ? true : false
 }
 
-String profile = isBuildingTag() ? '-Pprod' : ''
-
 pipeline {
     agent {
         node {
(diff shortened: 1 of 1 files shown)


Jenkinsfile
```diff
@@ -42,6 +41,21 @@ pipeline {
     }
 
     stages {
+
+        stage('Bump version and tag') {
+            when {
+                anyOf {
```
frisonisland: perhaps can be cleaned up, waiting for other people feedback.
I would run it always and minimize need of variables and conditions

Jenkinsfile
```diff
@@ -13,8 +13,7 @@ boolean isBuildingTag() {
     return env.TAG_NAME ? true : false
 }
 
-String profile = isBuildingTag() ? '-Pprod' :
-        (env.BRANCH_NAME == 'devel' ? '-Pdev' : '')
+String profile = isBuildingTag() ? '-Pprod' : ''
```
frisonisland: To be honest I would just build as production in CI at this point

keshavbhatt (approved):

PR #1107, opened 2026-07-17
test: kubernetes testcontainer support

frisonisland:
I didn't consider yet making a java library, but perhaps we should do it.

There are two classes than can be extracted in a library: KubernetesPods and the strategy pattern (ContainerRuntime).

Everything else is just an interface that implements two variations of the contracts: docker or kubernetes.
Every project should maintain their own, based on contract preferences and needs.

### DIFF
--- a/Jenkinsfile
+++ b/Jenkinsfile
@@ -71,6 +71,12 @@ pipeline {
         }
 
         stage('UT, IT') {
+            // Container-backed tests (e.g. Redis) spin up real pods through the in-cluster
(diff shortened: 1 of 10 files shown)


Jenkinsfile
```diff
@@ -71,6 +71,12 @@ pipeline {
         }
 
         stage('UT, IT') {
+            // Container-backed tests (e.g. Redis) spin up real pods through the in-cluster
```
frisonisland: drop comments

M0Rf30 (approved):

PR #1118, opened 2026-08-07
feat: [CO-4105] add count accounts by COS internal API

frisonisland:
## Why

`carbonio-license-service` needs to know how many accounts sit in a set of COSes, to decide whether an edition still has seats. It was reaching for the **admin SOAP API**, which meant two things we do not want:

- license-service would have to hold a **mailbox admin account and password** in its own config
- every seat check cost **1 + N SOAP calls** (one `SearchDirectory`, then one `CountAccount` per domain), from another process

Counting belongs where the data is. Inside mailboxd it is a local `Provisioning` call, needs no credentials, and collapses to a single request for the caller.

## What

`GET /internal/cos/count?cosId=…&cosId=…` → the account count per COS and their sum.

```json
{ "total": 15, "byCos": { "cos-a": 10, "cos-b": 5 } }
```

Also `GET /internal/cos/{id}`, so a caller can test COS existence without the admin API either — a `404` is absence.

The tally itself is unchanged: the same per-domain `CountAccount` the license counts with, so an account inheriting the domain or default COS is attributed to that COS rather than dropped. COS ids match case-insensitively, and a COS with no accounts comes back as `0` rather than being left out.

Both sit on `/internal/*`, which the existing `PortRestrictionFilter` already binds to the loopback internal-API port — no token to pass, nothing new to authenticate. `carbonio-mailbox-internal-api-intentions.json` gains `carbonio-license-service` as a source, since Connect would otherwise deny the call.

## Tests

`CosServiceCountTest` — 5 unit tests against a mocked `Provisioning`: summing one COS across domains, zero for a COS with no accounts, ignoring COSes that were not asked about, case-insensitive ids, and no `getAllDomains()` call at all for an empty request. Unit rather than the `@Tag("e2e")` style the other API tests use, since this is pure aggregation.

## Notes

- `store/internal-api-openapi.yaml` is regenerated by the build and included.
- Client side: `zextras/carbonio-mailbox-sdk` branch `feat/internal-api-cos-and-account-count`, adding `countCos` and `findCos`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


store/src/main/java/com/zextras/mailbox/api/rest/service/CosService.java
```diff
@@ -0,0 +1,78 @@
...
+      final Cos cos = provisioningSupplier.get().getCosById(cosId);
+      if (cos == null) {
+        throw ServiceException.NOT_FOUND("No such COS with ID: " + cosId);
+      }
+      return cos;
+    });
+  }
+
+  /**
+   * Counts the accounts in each of the given COSes, using the same per-domain tally the license
```
frisonisland: Ugly comments, no need to comment everything

keshavbhatt (approved):

PR #1141, opened 2026-09-15
feat(CLI): print quota sizes in a human readable format

gnekoz:
Refs: [CO-4266](https://zextras.atlassian.net/browse/CO-4266)

> [!IMPORTANT]
> Stacked on #1140. That PR is also open against `main` and this branch builds on its commit, so the diff below currently includes it. Once #1140 merges, this PR narrows to its own commit. Review the second commit only.

## Why

`zmprov gqu` and friends reported mailbox usage entirely in raw bytes, so administrators had to wrap them in `awk`/`perl`/bash helpers just to read the numbers for day-to-day management and reporting.

## What

`zmprov gqu`, `gmi` and `rmc` now print the largest fitting unit instead of the byte count:

```
$ zmprov gqu docker.carbonio.localhost
WARNING: This size refers to Mails/Calendars/Contacts only
zextras@carbonio.localhost 5.88 MB
user@carbonio.localhost 87.12 KB
admin@carbonio.localhost 0 B
WARNING: This size refers to Mails/Calendars/Contacts only
```

The formatting logic is **not new**. It is the private `ZMailboxUtil.formatSize`, extracted into `com.zextras.mailbox.util.ByteSizeFormatter` so both CLIs share one implementation. The unit convention (1024-based, `KB`/`MB`/`GB`, 2 decimals) is unchanged, so `zmmailbox` output is unaffected and `gms -v` still prints raw bytes as the machine readable escape hatch.

Two deliberate deltas while extracting:
- `>=` instead of `>` on the unit boundaries, so exactly 1 GiB prints `1.00 GB` rather than `1024.00 MB`.
- The `refeers` -> `refers` typo in the quota usage notice is fixed.

`rmc` is not named in the ticket but prints the same `quotaUsed` figure, so it is included for consistency.

## Notes for the reviewer

**Output format is now human readable only** — the raw byte count is replaced, not appended. Nothing in this repo parses `gqu`/`gmi` stdout, but ops tooling or CE scripts outside it might.

**No locale handling was added.** `String.format` follows the JVM default locale, so an Italian-locale server prints `5,88 MB`. This is pre-existing behaviour inherited from `formatSize` and was kept deliberately for consistency rather than pinned to `Locale.ROOT`.

## Testing

- New `ByteSizeFormatterTest` — 11 parameterized cases covering `0 B`, sub-KB, each unit boundary exactly, and `Long.MAX_VALUE`.
- `ProvUtilAPITest` `gmi`/`gqu` assertions updated; 70 tests pass.
- Verified end to end on the Docker stack against real mailbox data: `gqu`, `gmi`, `rmc`, plus `zmmailbox gms`, `gms -v` and the selectMailbox banner. Cross-checked the conversion: `gms -v` reports `6164825`, `6164825 / 1024² = 5.879` -> displayed `5.88 MB`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[CO-4266]: https://zextras.atlassian.net/browse/CO-4266?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


gnekoz:
## SonarQube quality gate

**1 New Issue — fixed** in d38c8d905a. It was `java:S6126` ("Replace this String concatenation with Text block") at `ProvUtilAPITest.java#L358`; the expected `gmi` output is now a text block. `ProvUtilAPITest` still passes (70 tests, 0 failures).

**Coverage on New Code (71.80% < 80%) — intentionally left as is.** Reproduced locally at 70.97%: 9 uncovered new lines.

| Where | Uncovered lines | Introduced by |
|---|---|---|
| `ZMailboxUtil` (select banner, `gms`, `dumpMessage`) | 7 | 4 from #1140, 3 from this PR |
| `ByteSizeFormatter` private constructor | 2 | this PR |

Two reasons for leaving it:

1. **Almost half the gap is not this PR's.** Sonar measures new code against the merge base with `main`, and #1140 sits in between, so its lines are counted here too. **Once #1140 merges, this PR's new code shrinks to its own commits and measures 80.00%, which clears the gate.**
2. **Covering `ZMailboxUtil` is not cheap.** No existing test boots the user-facing `/service/soap` endpoint that `ZMailbox` requires, and `ZMailboxUtil.stdout` is a `private static PrintWriter` with no setter, so a test needs reflection or a production visibility change. That is a legacy-code refactor beyond the scope of CO-4266.

The formatting logic itself — the only real logic added here — is fully covered by `ByteSizeFormatterTest` (11 cases). The uncovered lines are CLI plumbing.

Worth noting the margin after #1140 merges is exactly 80.00%, i.e. one line. If the team would rather have real headroom, the follow-up is a `zmmailbox` output test, which would also be the first regression test for that CLI's output.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

keshavbhatt (approved):

CerDennis (approved):

PR #1146, opened 2026-09-17
fix: [CO-4287] restore needrestart drop-in for mailboxd work dir

M0Rf30:
## Problem

The CO-3996 needrestart drop-in (38ac681) that suppresses needrestart's false-positive flag on `carbonio-appserver.service` only landed on the 26.6 retrofix branch (`co-3846/26.6`) and was never merged into `main`. As a result 26.9.0 ships without it: `zstd-jni` extracts its native lib into mailboxd's `java.io.tmpdir` and unlinks it while still mapped, so needrestart flags `carbonio-appserver.service` on every apt/dpkg transaction (CO-3996).

The OpenTelemetry eBPF (OBI) java agent introduces the same failure mode from a different angle: it unpacks nested jars into the JVM's `java.io.tmpdir` and deletes them after loading, which needrestart also detects.

## Fix

Restore and broaden the needrestart drop-in in `packages/appserver-service/`, installed to `/etc/needrestart/conf.d/carbonio-appserver.conf` from `_package()`:
- Extends `blacklist_mappings` to cover the whole mailboxd work dir, both the current `/opt/zextras/mailboxd/work` path and the legacy `/opt/zextras/jetty_base/work` path.
- Adds a matching `blacklist_interp` entry so needrestart also ignores the OBI-unpacked jars found via `/proc/PID/fd`.

## Verification

```
perl -e 'our %nrconf=(blacklist_mappings=>[],blacklist_interp=>[]); do "./packages/appserver-service/needrestart-carbonio-appserver.conf" or die $@; print scalar(@{$nrconf{blacklist_mappings}}), " ", scalar(@{$nrconf{blacklist_interp}}), "\n"'
1 1
```

Refs: https://zextras.atlassian.net/browse/CO-4287 , https://zextras.atlassian.net/browse/CO-3996

Companion PR in carbonio-core covers `/opt/zextras/data/tmp` for `carbonio-configd.service` and other `zmjava`-launched JVMs affected by the same OBI-agent behavior.


M0Rf30:
Behavioural check on `ubuntu:24.04` with needrestart `3.6-7ubuntu4.5`, one process per pattern (a python3 copy named `java` holding an unlinked jar; python3 processes with `PROT_EXEC` mappings of unlinked `.so` files), plus an uncovered control under `/opt/other/`:

```
=== WITHOUT drop-ins ===
[Core] #3149 uses obsolete script file(s):
[Core] #3149  /opt/zextras/data/tmp/obi-agent123/jarXYZ.jar (deleted)
[main] #3150 uses deleted /opt/zextras/mailboxd/work/libzstd-jni-1.5.6-123.so
[main] #3151 uses deleted /opt/zextras/jetty_base/work/libzstd-jni-1.5.6-456.so
[main] #3152 uses deleted /opt/other/libcontrol.so
=== WITH both CO-4287 drop-ins (carbonio-core.conf + carbonio-appserver.conf) ===
[Core] blacklisted: /opt/zextras/data/tmp/obi-agent123/jarXYZ.jar (deleted)
[main] #3152 uses deleted /opt/other/libcontrol.so
```

Only the control is still flagged; both mailboxd work-dir layouts and the OBI tmpdir jars are ignored.

keshavbhatt (approved):

